Issue #45, opened 2018-07-03
As a service Author I want to run acceptance tests using a suitable environment


ojhughes:
**Given** Environment secrets and credentials are available to the build
**When** When a build is triggered by a Pull Request in the open source project repository
**Then** The build uses a Cloud Foundry server to run acceptance tests against
**And** Does not expose any secrets in files or logs

csterwa:
@ojhughes is this in progress?

ojhughes:
where? (bosh-lite, BYOInstance, VM snapshot, PWS...)

Are ATs part of CircleCI or a separate thing?
Are ATs part of the same repo?
What are we going to use to code the ATs?
Are we going with the cf cli?

ojhughes:
- [ ] Document pros / cons for above questions

ojhughes:
Trying to create a service broker  within the AT setup but looks like CAPI cannot connect to the service broker:
```cf create-service-broker sample-broker user password https://sample-broker.mybroker.my-app.com
Creating service broker sample-broker as admin...
FAILED
Server error, status code: 502, error code: 10001, message: The service broker could not be reached: https://sample-broker.mybroker.my-app.com/v2/catalog```
I can curl the broker from my machine but not from cloud controller VM

PR #66, opened 2018-08-06
Add configuration to stop and start a GCP host BOSH Lite in the CircleCI pipeline

ojhughes:
Template credentials required by BOSH from Circle CI environment variables
Script to create CircleCI environments from file in LastPass
Ensure only a single job runs at a time so that there is not a deadlock over BOSH lite environment (in future multiple envs will be enabled)
Update Docker file to include bbl and bosh cli
[#45]


ci/config.yml
```diff
@@ -0,0 +1,85 @@
+# Java Gradle CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-java/ for more details
+#
+version: 2
+jobs:
+  build:
+    docker:
+      - image: ojhughes/spring-cloud-app-broker-ci:v1
```
AlbertoImpl: Can we use the account we have under `springcloudservices`?

ci/scripts/wait-for-other-builds-to-finish.sh
```diff
@@ -0,0 +1,42 @@
...
+			sleep 10
+			printf "."
+			if [[ "${SECONDS}" -gt "${max_time_seconds}" ]]; then
+				printf "Max time exceeded waiting to reach front of queue, cancelling build"
+				cancel_current_build
+				sleep 20
+				exit 1
+			fi
+	done
+	printf "Build is font of queue"
```
AlbertoImpl: r/font/front

ci/scripts/wait-for-other-builds-to-finish.sh
```diff
@@ -0,0 +1,42 @@
...
+	until [[ $(curl -s "https://circleci.com/api/v1.1/project/github/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}?circle-token=${CIRCLECI_API_KEY}&filter=running" \
+		| jq 'min_by(.build_num) | .build_num') \
+		-eq ${CIRCLE_BUILD_NUM} ]]; do
+
+			sleep 10
+			printf "."
+			if [[ "${SECONDS}" -gt "${max_time_seconds}" ]]; then
+				printf "Max time exceeded waiting to reach front of queue, cancelling build"
+				cancel_current_build
+				sleep 20
```
AlbertoImpl: why do we have to sleep here?

ci/scripts/populate-secrets-from-environment.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+set -peuo pipefail
```
AlbertoImpl: what is `-p` for?

ci/scripts/export-envvars-to-circle-ci.sh
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env bash
+set -euo pipefail
+readonly CIRCLE_API_BASE=https://circleci.com/api/v1.1/project/github/spring-cloud-incubator/spring-cloud-app-broker
+
+export_variables_to_circle() {
+	source "../.envrc"
+	#DIRECTOR_SSLCA
+	delete_envvar_from_circle DIRECTOR_SSLCA
```
AlbertoImpl: shall we not delete after the execution is done rather than on the next run?

ci/scripts/export-envvars-to-circle-ci.sh
```diff
@@ -0,0 +1,151 @@
...
+	if [[ -z "$DIRECTOR_SSLCA" ]]; then
+		echo "variable DIRECTOR_SSLCA must be set, make sure .envrc file is up to date"
+		exit 1
+	fi
+	create_envvar_in_circle DIRECTOR_SSLCA "${DIRECTOR_SSLCA}"
+
+	#JUMPBOX_PRIVATE_KEY
+	delete_envvar_from_circle JUMPBOX_PRIVATE_KEY
+	if [[ -z "$JUMPBOX_PRIVATE_KEY" ]]; then
```
AlbertoImpl: I wonder if we should be just checking for every var do be there first and then set them afterward so that we don't have some set and some not.

ci/scripts/populate-secrets-from-environment.sh
ojhughes: Typo! thanks

ci/scripts/export-envvars-to-circle-ci.sh
ojhughes: Agreed

ci/scripts/wait-for-other-builds-to-finish.sh
ojhughes: Just to give a chance for the build to be cancelled gracefully otherwise it would show as failed. I can reduce the wait time and add a comment

ci/scripts/wait-for-other-builds-to-finish.sh
ojhughes: Thanks!

ci/config.yml
ojhughes: Sure

ci/scripts/wait-for-other-builds-to-finish.sh
AlbertoImpl: I am not concerned about the time, but explaining the reason would be great. Thanks!

ci/scripts/export-envvars-to-circle-ci.sh
ojhughes: Actually this script is run manually using environment vars in last pass. I will add a comment explaining this. I'm not sure we need to delete at all so can remove the delete statements to avoid confusion

ci/config.yml
```diff
@@ -0,0 +1,85 @@
+# Java Gradle CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-java/ for more details
+#
+version: 2
+jobs:
+  build:
+    docker:
+      - image: springcloudservices/spring-cloud-app-broker-ci:v1
```
AlbertoImpl: As Scott pointed out, might be best to create a new account that it's just for OSS rather than mixing. In case we want to grant access to members outside of the team. Happy to have it in a separate chore.

ci/scripts/export-envvars-to-circle-ci.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+set -euo pipefail
+readonly CIRCLE_API_BASE=https://circleci.com/api/v1.1/project/github/spring-cloud-incubator/spring-cloud-app-broker
+readonly PROG_DIR=$(readlink -m $(dirname $0))
+readonly CI_DIR=$(dirname "${PROG_DIR}")
+
+export_variables_to_circle() {
+	[[ ! -f "${CI_DIR}/.envrc" ]] && printf "Download .envrc from LastPass using \"lpass show --notes 7083089362665788436\" in directory ${SCRIPTS_DIR} before running script" && exit 1
+#	source "$CI_DIR/.envrc"
```
AlbertoImpl: Do we want to have this here?

ci/scripts/wait-for-other-builds-to-finish.sh
```diff
@@ -0,0 +1,44 @@
...
+	echo "Must provide Max Queue Time in *minutes* as script argument"
+	exit 1
+fi
+
+readonly MAX_TIME=$1
+
+wait_for_previous_builds_to_complete(){
+	local max_time_seconds=$((MAX_TIME * 60))
+
+	until [[ $(curl -s "https://circleci.com/api/v1.1/project/github/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}?circle-token=${CIRCLECI_API_KEY}&filter=running" \
```
AlbertoImpl: So, locally:
```
Using build environment variables:
  BASH_ENV=/tmp/.bash_env-localbuild-1533734475
  CI=true
  CIRCLECI=true
  CIRCLE_BRANCH=oh-bosh
  CIRCLE_BUILD_NUM=
  CIRCLE_JOB=build
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
```
`CIRCLE_BUILD_NUM` is not set, we should make the script noop if that is the case

ci/config.yml
scottfrederick: There is a `springcloud` repo on Docker Hub. We should be able to get access to that. https://hub.docker.com/r/springcloud/

AlbertoImpl:
The build is passing.
I doubled checked the logs and looks good, no creds anywhere.
I was able to successfully run the ATs locally using `circleci` CLI.
👍

Issue #20, opened 2018-05-03
As a service author, I want service instance creation parameters mapped to SPRING_APPLICATION_JSON on deployed applications


royclarkson:
`GIVEN` that I specify a backing app with the name: `mapping-app`
`WHEN` `cf create-service my-service -c '{"key": "value", "compoundKey": { "subkey1": "subvalue1", "subkey2": "subvalue2" }}'`
`THEN` I can see the SPRING_APPLICATION_JSON environment variable for the backing app `mapping-app` containing the value:

```
"{"transformed.key": "value", "other.transformed.compoundKey"; { "subkey1": "subvalue1", "subkey2": "subvalue2" }}"
```
 where the mapping of keys from the request parameters to backing app env is configurable by the implementing service broker

PR #75, opened 2018-08-22
Adding environment variable from create-service params

AlbertoImpl:
Finishes #20


AlbertoImpl:
Some trade-offs:
The `ParametersTransformer` interface takes all the backing apps because otherwise, we wouldn't allow users to config properties for each one of them.
I went for the simple solution of adding the `Bean` instead of going all the way through strategy matched via config.

PR #84, opened 2018-08-29
Allow parameters transformers to be selectable by backing app.

scottfrederick:
This change allows parameters transformer implementations to be selected by configuration for each backing application: 

```
spring:
  cloud:
    appbroker:
      services:
        - service-name: example
          plan-name: standard
          apps:
            - name: broker-sample-app1
              path: classpath:demo.jar
              parameters-transformers: 
              - SimpleMapping
              - CustomMapping
            - name: broker-sample-app2
              path: classpath:demo.jar
              parameters-transformers: 
              - SomeOtherCustomMapping
```

Where the name `SimpleMapping` maps to a bean in the application context named `SimpleMappingParametersTransformer` that implements a `ParametersTransformer` interface. This project will have a set of transformer implementations available out-of-the-box for brokers to use, and implementing projects can contribute their own custom implementations.


spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/extensions/parameters/ParametersTransformationService.java
```diff
@@ -0,0 +1,73 @@
...
+
+	public Mono<BackingApplications> transformParameters(BackingApplications backingApplications,
+														 Map<String, Object> parameters) {
+		return Flux.fromIterable(backingApplications)
+			.flatMap(backingApplication ->
+				Flux.fromIterable(getTransformers(backingApplication))
+					.flatMap(transformerName ->
+						transform(transformerName, backingApplication, parameters)
+					))
+			.then(Mono.just(backingApplications));
```
royclarkson: I have a slight concern about this. It might be preferable to collect the backing apps and return that collection instead of the original reference to the `backingApplications` object. I think it should be ok though if it's not operating on a different thread.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/extensions/parameters/ParametersTransformationService.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private Mono<BackingApplication> transform(String transformerName,
+											   BackingApplication backingApplication,
+											   Map<String, Object> parameters) {
+		return getTransformerByName(transformerName, backingApplication.getName())
+			.flatMap(transformer -> transformer.transform(backingApplication, parameters));
+	}
+
+	private Mono<ParametersTransformer> getTransformerByName(String transformerName, String applicationName) {
+		return Mono.defer(() -> {
```
royclarkson: I like this

spring-cloud-app-broker-sample/src/test/java/org.springframework.cloud.appbroker/sample/CreateInstanceWithCreationParametersComponentTest.java
```diff
@@ -38,7 +38,8 @@
 	"spring.cloud.appbroker.services[0].service-name=example",
 	"spring.cloud.appbroker.services[0].plan-name=standard",
 	"spring.cloud.appbroker.services[0].apps[0].path=classpath:demo.jar",
-	"spring.cloud.appbroker.services[0].apps[0].name=" + APP_NAME
+	"spring.cloud.appbroker.services[0].apps[0].name=" + APP_NAME,
+	"spring.cloud.appbroker.services[0].apps[0].parameters-transformers[0]=SimpleMapping"
```
royclarkson: really feels like this will be useful for implementing brokers 👍

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/extensions/parameters/ParametersTransformationService.java
scottfrederick: Returning a `Mono<List<BackingApplication>>` instead of `Mono<BackingApplications>` is problematic. That would mean part of the flow deals with `BackingApplications` and the rest of the flow deals with `List<BackingApplication>`. That inconsistency is going to be confusing. 

We might be able to start the flow with `List<BackingApplication>` and be consistent with that throughout. Then `BackingApplications` would only be used for the Boot properties mapping. I can try that out.

PR #100, opened 2018-09-12
Implement service updates

spikymonkey:
This is a really naive implementation of update. It's almost a straight
copy of the create workflow, so using it will cause app downtime. It's
not intended to be a final implementation, just something to fill the
gap until we can do this properly.


scottfrederick:
@spikymonkey One thing to note: this project uses tabs for indentation, per the Spring project convention. Please configure your IDE to use tabs for contributions.

spikymonkey:
Ah sorry about that!

PR #104, opened 2018-09-25
Run service workflows sequentially

alek-sys:
Having `@Order` kind of implies sequential execution of update / create / delete flows. However with `flatMap` if lower order workflow completes first it will take precedence. `concatMap` will run the next flow only when the previous one is completed. 

Not sure if that should be a default behaviour. If we want some flexibility in here, we might need to provide some way to configure this behaviour.


scottfrederick:
@alek-sys I had intended the default behavior to be running the workflows sequentially. I think this is a good default, we can add other options later if someone has a need.

PR #108, opened 2018-10-03
Hide potentially sensitive logs output when serializing backing app

alek-sys:
Closes https://github.com/spring-cloud-incubator/spring-cloud-app-broker/issues/107


scottfrederick:
Looks great - thanks @alek-sys!

PR #120, opened 2018-10-10
Support authentication to CF via OAuth.

scottfrederick:
WIP - do not merge


build.gradle
```diff
@@ -117,8 +117,8 @@ configure(allprojects) {
 		"-Xlint:rawtypes",
 		"-Xlint:deprecation",
 		"-Xlint:unchecked",
-		"-Xlint:-options",
-		"-Werror"
+		"-Xlint:-options"
+//		"-Werror"
```
AlbertoImpl: Had the same problem when upgrading to the new CFJavaClient, kind of annoying that we have so many warnings. That is something to check with them.

PR #135, opened 2018-10-23
fix group id and exclude subprojects from publishing and apidocs

### LINKED ISSUE #110: Fix maven coordinates and filter artifacts
Artifacts are being published to the Spring repo with the group ID `org.springframework.cloud.appbroker`. This should be just `org.springframework.cloud`. 

The modules `spring-cloud-app-broker-sample` and `spring-cloud-app-broker-acceptance-tests` are being published along with the library artifacts, but they should be excluded from publishing.

LittleBaiBai:
Resolves #110

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ buildscript {
 }
 
 configure(allprojects) {
-	group = "org.springframework.cloud.appbroker"
+	group = "org.springframework.cloud"
 
 	apply plugin: "java-library"
 	apply plugin: "maven"
@@ -286,6 +286,7 @@ configure(rootProject) {
 			project.sourceSets.main.compileClasspath
 		})
 
+		exclude 'org/springframework/cloud/appbroker/sample/**'
 		maxMemory = "1024m"
 		destinationDir = new File(buildDir, "api")
 	}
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -286,6 +286,7 @@ configure(rootProject) {
 			project.sourceSets.main.compileClasspath
 		})
 
+		exclude '**/sample/**'
```
royclarkson: I wonder if we shouldn't narrow it down to the sample app package? Would this work?
```suggestion
		exclude 'org.springframework.cloud.appbroker.sample**'
```

LittleBaiBai: That would work. I was thinking this configuration would be applicable for other project with sample app as well. Is it common to have production code reside in a `sample` package?

royclarkson: We shouldn't have to worry about other production sample apps, since those shouldn't be published dependencies (artifacts).

LittleBaiBai:
@royclarkson moved to pull request #136

build.gradle
LittleBaiBai: @royclarkson applied your comment in pull request #136

Issue #139, opened 2018-11-05
As service author, I can specify in a manifest the strategy for additional parameters in the backing service


pivotal-dylan-roberts:
`AS` service author, I can specify in a manifest the strategy for additional parameters in the backing service that gets deployed to CF on service create
`SO THAT` I can customize the backing services parameters dynamically.

`GIVEN` the application manifest:
```
spring:
  cloud
    appbroker:
      services: 
        apps:
        - name: app-with-services
           services:
           - service-instance-name: my-db-service
            parameters-transformers:
             - name: ParameterMapping
               include: paramA,paramC
        services:
        - service-instance-name: my-db-service
          name: mysqldb
          plan: small
```
`WHEN` a service gets created with `cf create-service my-service -c '{"paramA":"valueA", "paramB":"valueB", "paramC":"valueC"}'`
`THEN` the SI `my-db-service` will be created with the parameters `-c '{"paramA":"valueA", "paramC":"valueC"}'`

scottfrederick:
Depends on #14

scottfrederick:
@pivotal-dylan-roberts Does your use case require that the parameters the user provides in the create-service parameters be transformed in any way before being provided to the backing service instance (e.g. changing keys provided in the JSON hash), or can the parameter keys and values be copied as-is?

AlbertoImpl:
Will this match your expection @pivotal-dylan-roberts ? 

`GIVEN`
```
spring:
  cloud
    appbroker:
      services: 
        apps:
        - name: app-with-services
           services:
           - service-instance-name: my-db-service
             service-parameters-transformers: InputApp1Transformer
        services:
        - service-instance-name: my-db-service
          name: mysqldb
          plan: small
```
`AND` `InputApp1Transformer` is specified to transform a value A into a value B
`WHEN` `cf create-service my-service -c '{"paramA":"valueA"}'`
`THEN` the SI `my-db-service` will be created with `-c '{"paramA":"valueB"}'`


Feel free to update it, specially the `THEN` clause since we can't really validate it.

Also, as Scott mentioned, would this suffice as a first story?
`GIVEN`
```
spring:
  cloud
    appbroker:
      services: 
        apps:
        - name: app-with-services
           services:
           - service-instance-name: my-db-service
        services:
        - service-instance-name: my-db-service
          name: mysqldb
          plan: small
```
`WHEN` `cf create-service my-service -c '{"paramA":"valueA"}'`
[...]

PR #147, opened 2018-11-12
Adding parameter transformer to backing services

AlbertoImpl:
Generalizing ParameterTransformers
Attempting to clean up before each test
Adding support for nested objects structure as parameters
Adding version to the mocked get-info response

Finishes #139


scottfrederick:
Nice work!

Issue #17, opened 2018-05-03
As a service author, I can describe in a manifest that OAuth2 credentials should be generated for a deployed application


royclarkson:
`GIVEN` The application manifest contains:

```
spring:
  cloud:
    appbroker: 
      services: 
        apps:
        - name: my-app
          credentials-providers:
          - name: SpringSecurityOAuth2
            args: 
            - identity-zone-subdomain: my-identity-zone
```

`WHEN` a service instance is created
`THEN` the application `my-app` has a `SPRING_APPLICATION_JSON` environment variable containing the values:

```
{
"security.oauth2.client.access-token-uri: "https://my-identity-zone.uaa.example.com/oauth/token", 
"security.oauth2.client.client-id": "my-app-[SI GUID]", 
"security.oauth2.client.client-secret": "a-generated-secret"
}
```

`AND` a client with the ID and secret will be stored in UAA within the specified identity zone
`AND` the client will be deleted from UAA when the service instance is deleted

If the `args: identity-zone-subdomain` configuration property is omitted, then the client is created in the default UAA identity zone.

scottfrederick:
I believe this story will also require that the generated client ID and secret are stored in the OAuth2 provider when they are generated. If so, the story should be specific to CF UAA, and use the UAA REST API to store the generated ID and secret.

PR #154, opened 2018-11-27
Add a credential provider implementation for OAuth2 clients

scottfrederick:
Finishes #17


scottfrederick:
@AlbertoImpl Good idea. The core will be to run ATs in CI using an OAuth2 client for authentication to CF (instead of a username/password) with some specific authorities on the client. OAuth2 authentication is something we should be testing anyway, but it will require some setup on the target CF env after its deployed.

PR #155, opened 2018-11-28
Gradle cleanup

royclarkson:


spring-cloud-app-broker-deployer-cloudfoundry/build.gradle
```diff
@@ -16,29 +16,28 @@
 
 description = "Spring Cloud App Broker Deployer Cloud Foundry"
 
-ext {
-	reactorCoreVersion = '3.2.0.RELEASE'
-	reactorNettyVersion = '0.8.0.RELEASE'
-	cfJavaClientVersion = '3.13.0.RELEASE'
+dependencyManagement {
+	imports {
+		mavenBom "org.springframework.boot:spring-boot-dependencies:${springBootVersion}"
```
scottfrederick: Should the `-deployer-cloudfoundry` module depend on Boot?

scottfrederick: On further review, I don't think there's any harm in using Boot dependencies, and we could probably have the `-core` module depend on Boot also.

PR #165, opened 2018-12-07
Store binding credentials in CredHub when available

### LINKED ISSUE #119: As a service author, I want binding credentials provided by the service broker to be stored in CredHub
GIVEN a service broker is deployed using the manifest: 
```
spring:
  cloud:
    appbroker: 
      services:
        service-name: service-with-custom-binding-handler
        plan-name: standard 
        apps:
        - name: instance-to-bind-with
          path: classpath:demo.jar 
```
AND a service instance is provisioned
WHEN a `bind-service` request is received from the platform
THEN the `credentials` field of the `bind-service` response is populated by the implementing broker
AND the  credentials are stored in CredHub using the `cloud-foundry-extensions` module
AND the `credentials` field is populated with CredHub reference

CredHub credential creation will be conditional on the presence of a configure `CredhubTemplate`

royclarkson:
Resolves #119

### DIFF
--- a/spring-cloud-app-broker-autoconfigure/src/main/java/org/springframework/cloud/appbroker/autoconfigure/AppBrokerAutoConfiguration.java
+++ b/spring-cloud-app-broker-autoconfigure/src/main/java/org/springframework/cloud/appbroker/autoconfigure/AppBrokerAutoConfiguration.java
@@ -211,13 +211,13 @@ public class AppBrokerAutoConfiguration {
 
 	@Bean
 	public WorkflowServiceInstanceBindingService serviceInstanceBindingService(
-			ServiceInstanceBindingStateRepository stateRepository,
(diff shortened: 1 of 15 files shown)


spring-cloud-app-broker-autoconfigure/src/main/java/org/springframework/cloud/appbroker/autoconfigure/AppBrokerAutoConfiguration.java
```diff
@@ -71,6 +75,9 @@
 
 	private static final String PROPERTY_PREFIX = "spring.cloud.appbroker";
 
+	@Value("${spring.application.name:unknown}")
```
AlbertoImpl: Shouldn't we fail-fast instead of providing a default value?

royclarkson: @AlbertoImpl I'm fine with that. I'll remove the default. Thanks.

PR #166, opened 2018-12-10
Add weekly job to cleanup bosh blobstore to avoid running out of disk space

ojhughes:


ojhughes:
Don't merge yet, seems to have stopped CI running tests on PR

PR #167, opened 2018-12-11
 Update backing service instances on update service.

### LINKED ISSUE #163: As service author, I can describe in a manifest services that get updated when a service instance is updated
`GIVEN` The application manifest contains:
```
spring:
  cloud
    appbroker:
      services: 
        apps:
        - name: app-with-services
          services:
          - service-instance-name: my-db-service
            name: mysqldb
            plan: small
          - service-instance-name: my-messaging-service
            name: rabbitmq
            plan: medium
```
`WHEN` a service instance is updated with `cf update-service -c '{"key":"value"}'`
`THEN` backing service instances `my-db-service` and `my-messaging-service` are updated (asynchronously if required by the managing service broker) with appropriate parameter mapping (as supported by #139) 

**Note:** services instances can be updated with a change to the SI name, plan, or parameters. Among these attributes, only backing service instance parameters can potentially be affected by an update to the managed service instance. An update to the managed SI will only result in an update of a backing SI if the mapping of incoming parameters (from e.g. `cf update-service -c`) to backing SI parameters results in parameters being configured for the backing SI.

scottfrederick:
Fixes #163

### DIFF
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org.springframework.cloud.appbroker.acceptance/UpdateInstanceWithServicesAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org.springframework.cloud.appbroker.acceptance/UpdateInstanceWithServicesAcceptanceTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2016-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.appbroker.acceptance;
+
+import org.cloudfoundry.operations.applications.ApplicationSummary;
+import org.cloudfoundry.operations.services.ServiceInstanceSummary;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class UpdateInstanceWithServicesAcceptanceTest extends CloudFoundryAcceptanceTest {
+
+	private static final String APP_NAME = "app-update-services";
+	private static final String SI_NAME = "si-update-services";
+
+	private static final String BACKING_SERVICE_NAME = "backing-service-update";
+	private static final String BACKING_SI_NAME = "backing-service-instance-update";
+
+	@BeforeEach
+	void setUpServiceBrokerForService() {
+		deployServiceBrokerForService(BACKING_SERVICE_NAME);
+	}
+
+	@AfterEach
+	void tearDownServiceBrokerForService() {
+		deleteServiceBrokerForService(BACKING_SERVICE_NAME);
+	}
+
+	@Test
+	@AppBrokerTestProperties({
+		"spring.cloud.appbroker.services[0].service-name=example",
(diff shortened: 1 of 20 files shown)


spring-cloud-app-broker-acceptance-tests/src/test/java/org.springframework.cloud.appbroker.acceptance/UpdateInstanceWithServicesAcceptanceTest.java
```diff
@@ -0,0 +1,89 @@
...
+		Optional<ServiceInstanceSummary> serviceInstance = getServiceInstance(SI_NAME);
+		assertThat(serviceInstance).hasValueSatisfying(value ->
+			assertThat(value.getLastOperation()).contains("completed"));
+
+		// then a backing application is deployed
+		Optional<ApplicationSummary> backingApplication = getApplicationSummaryByName(APP_NAME);
+		assertThat(backingApplication).hasValueSatisfying(app ->
+			assertThat(app.getRunningInstances()).isEqualTo(1));
+
+		Map<String, Object> parameters = new HashMap<>();
```
AlbertoImpl: Silly thing, but I had to read the code and the comments twice in order to understand it.
I would go with a: 
```
// given a SI is created 
// and a backing app is deployed 
// when I update the service instance  
//then the services are bound to it 
// and the SI can be deleted
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org.springframework.cloud.appbroker.acceptance/UpdateInstanceWithServicesAcceptanceTest.java
```diff
@@ -0,0 +1,89 @@
...
+		parameters.put("parameter1", "value1");
+		parameters.put("parameter2", "value2");
+		parameters.put("parameter3", "value3");
+		updateServiceInstance(SI_NAME, parameters);
+
+		// the services are bound to it
+		Optional<ServiceInstanceSummary> serviceInstance1 = getServiceInstance(BACKING_SI_NAME);
+		assertThat(serviceInstance1).hasValueSatisfying(instance ->
+			assertThat(instance.getApplications()).contains(APP_NAME));
+
```
AlbertoImpl: Shouldn't we assert that the service was updated?

spring-cloud-app-broker-sample/src/test/java/org.springframework.cloud.appbroker/sample/UpdateInstanceWithServicesComponentTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+	@Test
+	void updateAppWithServicesWhenServicesExist() {
+		cloudControllerFixture.stubAppExists(APP_NAME);
+		cloudControllerFixture.stubUpdateApp(APP_NAME);
+
+		cloudControllerFixture.stubServiceInstanceExists(SERVICE_INSTANCE_NAME);
+		cloudControllerFixture.stubCreateServiceBinding(APP_NAME, SERVICE_INSTANCE_NAME);
+
+		// when a service instance is created
```
AlbertoImpl: created -> updated

spring-cloud-app-broker-acceptance-tests/src/test/java/org.springframework.cloud.appbroker.acceptance/UpdateInstanceWithServicesAcceptanceTest.java
scottfrederick: I couldn't figure out how to do that, since there's no before/after properties to query on the backing service. I'll think about it more. 

I'm planning to remove pushing a second app broker and using that for a backing SI broker, as pushing that that second broker app and then having it push additional backing apps takes a lot of time. So I don't want to rely on that mechanism for checking before/after state of a backing SI.

spring-cloud-app-broker-sample/src/test/java/org.springframework.cloud.appbroker/sample/UpdateInstanceWithServicesComponentTest.java
scottfrederick: This is the biggest problem I have with this style of `given`/`when`/`then` comments in these tests - they're just comments and the don't keep up with the code. They can help with readability, but they can also lie. I've seen mistakes like this in several of the AT's I've been working on. I'll fix this one.

PR #168, opened 2018-12-11
Add support to optionally generate secrets from CredHub

### LINKED ISSUE #152: As a service author, I want any credentials or secrets to be generated by CredHub
`GIVEN` a service broker is using one of the credentials providers provided by App Broker in #16 or #17 
`WHEN` a service instance is provisioned
`THEN` CredHub is used to generate and store any credentials or secrets
`AND` when a service instance is deleted
`THEN` the generated credentials or secrets are deleted from CredHub

royclarkson:
- Add `CredHubCredentialsGenerator`
- Update `CredHubAutoconfiguration` to conditionally create beans
- Refactor the `CredentialGenerator` interface to be reactive
- Replace use of Pair in favor of Tuple2

Resolves #152

### DIFF
--- a/spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/credentials/SimpleCredentialGeneratorTest.java
+++ b/spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/credentials/SimpleCredentialGeneratorTest.java
@@ -16,8 +16,8 @@
 
 package org.springframework.cloud.appbroker.extensions.credentials;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.junit.jupiter.api.Test;
+import reactor.test.StepVerifier;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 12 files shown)


spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/credentials/SimpleCredentialGeneratorTest.java
```diff
@@ -27,28 +27,34 @@
 	void generateString() {
 		SimpleCredentialGenerator generator = new SimpleCredentialGenerator();
 
-		assertThat(generator.generateString(null, null, 12, true, false, false, false))
+		assertThat(generator.generateString(null, null, null, 12, true, false, false, false)
```
scottfrederick: This could use `StepVerifier.create().assertNext()` instead of blocking.

spring-cloud-app-broker-security-credhub/src/main/java/org/springframework/cloud/appbroker/extensions/credentials/CredHubCredentialsGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+		return credHubOperations.credentials().deleteByName(credentialName(applicationId, serviceInstanceId, descriptor));
+	}
+
+	@Override
+	public Mono<Void> deleteString(String applicationId, String serviceInstanceId, String descriptor) {
+		return credHubOperations.credentials().deleteByName(credentialName(applicationId, serviceInstanceId, descriptor));
+	}
+
+	private SimpleCredentialName credentialName(String applicationId, String serviceInstanceId, String descriptor) {
+		return new SimpleCredentialName(applicationId + "-" + serviceInstanceId + "-" + descriptor);
```
scottfrederick: This will generate a single-level name with dashes instead of the path-like structure. I think `new SimpleCredentialName(applicationId, serviceInstanceId, descriptor)` would be better.

PR #169, opened 2018-12-11
Create a spring-cloud-starter-app-broker-cloudfoundry and move all CF-specific dependencies.

scottfrederick:
This allows an app (like the `-sample` app) to only depend on two starters - a base App Broker starter and a platform-specific App Broker starter - and get all other dependency versions managed.

```
dependencies {
	implementation "org.springframework.cloud:spring-cloud-starter-app-broker"
	implementation "org.springframework.cloud:spring-cloud-starter-app-broker-cloudfoundry"

...
}
```


spring-cloud-starter-app-broker-cloudfoundry/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+
+dependencies {
+	api project(":spring-cloud-starter-app-broker")
+	api project(":spring-cloud-app-broker-deployer-cloudfoundry")
+	api project(":spring-cloud-app-broker-security-credhub")
+
+	api("org.cloudfoundry:cloudfoundry-client-reactor:${cfJavaClientVersion}")
+	api("org.cloudfoundry:cloudfoundry-operations:${cfJavaClientVersion}")
+	api("io.projectreactor.ipc:reactor-netty")
+	api("org.immutables:value:${immutablesVersion}")
```
royclarkson: I don't think we should include this immutables dependency in the starter. I added it to the other build files to clean up some Gradle warnings, but it should really be handled by Reactor. And maybe it's actually a bug that we have to include it at all?

spring-cloud-app-broker-core/build.gradle
```diff
@@ -29,13 +29,14 @@ dependencyManagement {
 }
 
 dependencies {
-	implementation project(":spring-cloud-app-broker-deployer")
+	api project(":spring-cloud-app-broker-deployer")
+	api("io.projectreactor:reactor-core")
+	api("org.apache.commons:commons-text:1.4")
+	api("org.springframework.cloud:spring-cloud-open-service-broker-core:${openServiceBrokerVersion}")
```
royclarkson: We should probably be using `spring-cloud-starter-open-service-broker` here. Also, this dependency is duplicated below as an `implementation` dependency.

spring-cloud-starter-app-broker-cloudfoundry/build.gradle
scottfrederick: I'll try removing that and seeing what happens with the build and warnings. I'm pretty sure it isn't needed at runtime.

spring-cloud-app-broker-core/build.gradle
scottfrederick: I like the idea of using `spring-cloud-open-service-broker-core` as the minimum (transitive) dependency in `spring-cloud-app-broker-core` and having the app broker starter depend on the `spring-cloud-starter-open-service-broker` (as things are now). This makes extensions like `spring-cloud-app-broker-security-credhub` cleaner, in that they can depend on `-core` and only get necessary dependencies without also getting auto-config etc.

PR #175, opened 2018-12-13
Upgrading Spring Boot to 2.1.1

AlbertoImpl:
Upgrading reactor and core
Pinning assertJ version

Finishes #160

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,14 +44,15 @@ configure(allprojects) {
 	apply plugin: "com.jfrog.artifactory"
 
 	ext {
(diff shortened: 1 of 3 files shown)


scottfrederick:
@arios @royclarkson I'm surprised that changing Reactor version from `Bismuth` to `Californium` doesn't require all references to the artifact `io.projectreactor.ipc:reactor-netty` to be changed to `io.projectreactor.netty:reactor-netty`, as this coordinate name change was made between those Reactor release trains.

PR #180, opened 2018-12-20
Use qualifiers when setting up cf deployer beans

ojhughes:
This allows consuming libraries to provide beans of the same type without conflict


ojhughes:
Thanks both!

PR #186, opened 2019-01-08
Add dir and modify build task for reference docs

fifthposition:
`./gradlew docs` builds documentation in `spring-cloud-app-broker-docs/build/`.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.8.1-bin.zip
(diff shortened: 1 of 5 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.8.1-bin.zip
```
scottfrederick: Is there something that requires the Gradle version downgrade?

fifthposition: I think I re-ran `gradle wrapper` or something. Certainly I don't remember modifying this file by hand. Will revert.

PR #200, opened 2019-02-01
First draft of reference documentation

fifthposition:
Fixes #111

PR #207, opened 2019-02-26
Updates to reference documentation

fifthposition:
From feedback on #200. Thanks @scottfrederick!

### DIFF
--- a/spring-cloud-app-broker-docs/src/docs/asciidoc/deployment-platforms.adoc
+++ b/spring-cloud-app-broker-docs/src/docs/asciidoc/deployment-platforms.adoc
@@ -16,6 +16,10 @@ spring:
           *api-port: 443*
           *username: admin*
           *password: adminpass*
+          *client-id: EXAMPLE_ID*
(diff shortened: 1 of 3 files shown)


spring-cloud-app-broker-docs/src/docs/asciidoc/deployment-platforms.adoc
```diff
@@ -16,6 +16,10 @@ spring:
           *api-port: 443*
           *username: admin*
           *password: adminpass*
+          *client-id: EXAMPLE_ID*
+          *client-secret: EXAMPLE_SECRET*
           *default-org: test*
           *default-space: development*
 ```
+
+NOTE: The two properties `username` and `password` and the two properties `client-id` and `client-secret` are mutually exclusive. The `client-id` and `client-secret` properties are for use with OAuth 2.0.
```
fifthposition: I don’t know if I like this way of presenting it, but short of adding a separate config section example I couldn't think of how else to do it--LMK if we should just have two examples to minimize confusion.

PR #258, opened 2019-07-31
Fix blocking calls to workflow layers ...

### LINKED ISSUE #256: Blocking requests during instance creation call
While working on a fix for #254 I've found a different flaw that looks a bit odd to me (but maybe there is some reason that it is how it is):
After creating the response for a service instance creation, the execution is switched to a parallel scheduler:
https://github.com/spring-cloud/spring-cloud-app-broker/blob/12d1a5b764cedd67e6579531f7577346f9bc88cc/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceService.java#L79
The thing is, that even though the execution is now running on a different thread, the return of the response is still blocked and not yet happening as the parallel executor thread is first running the subscription in the following `doOnNext` method.
From my point of view the `.publishOn(Schedulers.parallel())` should be moved before this line:
https://github.com/spring-cloud/spring-cloud-app-broker/blob/12d1a5b764cedd67e6579531f7577346f9bc88cc/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceService.java#L100
By that the execution would save the state in the current request thread, switch the following execution to the parallel thread and would return the response on the request thread.
This would make much more sense to me. What do you think?

EDIT: This is of course also true for updating and deleting a service instance and for the create and delete method in the `WorkflowServiceInstanceBindingService` class.

GFriedrich:
…and return a response as soon as possible

Resolves #256

The `sleep` calls within the mono should simulate a long running computation within the workflow layer. If you revert the changes in the implementation, the tests fail.
I know the solution is looking odd with all the `sleep` calls. I'm aware that such kind of tests tend to become flaky. Unfortunately I don't have a better idea how to handle this, as the execution of the workflows is now really async.
In case you've a better idea, how to test this, feel free to change. :)

### DIFF
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
@@ -80,7 +80,6 @@ public class WorkflowServiceInstanceBindingService implements ServiceInstanceBin
 	@Override
 	public Mono<CreateServiceInstanceBindingResponse> createServiceInstanceBinding(CreateServiceInstanceBindingRequest request) {
 		return invokeCreateResponseBuilders(request)
-			.publishOn(Schedulers.parallel())
 			.doOnNext(response -> create(request, response)
 				.subscribe());
 	}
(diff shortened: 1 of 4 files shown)


GFriedrich:
@Albertoimpl Any objections on merging this one too? Or do you need some time to think about those tests?

GFriedrich:
@Albertoimpl another month later this PR is still open - is there anything I can do to have this PR merged?

royclarkson:
@GFriedrich we haven't forgotten about this. Thanks for the reminder.

royclarkson:
Closed in 9072f858361be9a2acc8edae659ad98b309a801c

PR #270, opened 2019-09-11
Upgrading dependencies

Albertoimpl:
Upgrading dependencies
Deleting bom overrides
Deleting spring platform references
Removing spring.io plugin 
Unpinning jetty version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,6 @@ buildscript {
 	dependencies {
 		classpath("io.spring.gradle:propdeps-plugin:0.0.10.RELEASE")
 		classpath("io.spring.gradle:dependency-management-plugin:1.0.8.RELEASE")
-		classpath("io.spring.gradle:spring-io-plugin:0.0.8.RELEASE")
 		classpath("org.asciidoctor:asciidoctor-gradle-plugin:1.6.1")
 		classpath("org.jfrog.buildinfo:build-info-extractor-gradle:4.9.9")
 		classpath("io.spring.nohttp:nohttp-gradle:0.0.3.RELEASE")
--- a/build.gradle
+++ b/build.gradle
@@ -20,11 +20,11 @@ buildscript {
 	}
 	dependencies {
 		classpath("io.spring.gradle:propdeps-plugin:0.0.10.RELEASE")
-		classpath("io.spring.gradle:dependency-management-plugin:1.0.5.RELEASE")
+		classpath("io.spring.gradle:dependency-management-plugin:1.0.8.RELEASE")
 		classpath("io.spring.gradle:spring-io-plugin:0.0.8.RELEASE")
-		classpath("org.asciidoctor:asciidoctor-gradle-plugin:1.5.3")
-		classpath("org.jfrog.buildinfo:build-info-extractor-gradle:4.7.5")
-		classpath("io.spring.nohttp:nohttp-gradle:0.0.1.RELEASE")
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -54,7 +54,7 @@ configure(allprojects) {
 		springBootVersion = project.findProperty("springBootVersion") ?: "2.1.4.RELEASE"
 		springVersion = project.findProperty("springVersion") ?: "5.1.6.RELEASE"
 		reactorVersion = project.findProperty("reactorVersion") ?: "Californium-SR6"
-		openServiceBrokerVersion = "3.0.0.RELEASE"
+		openServiceBrokerVersion = "3.0.3.RELEASE"
```
royclarkson: I just published 3.0.4.RELEASE. Mind bumping this?

Albertoimpl: Just force pushed it. Thanks for the release @royclarkson!

build.gradle
```diff
@@ -54,7 +54,7 @@ configure(allprojects) {
 		springBootVersion = project.findProperty("springBootVersion") ?: "2.1.4.RELEASE"
```
royclarkson: Should we upgrade the boot version to 2.1.8.RELEASE? Also Spring Framework version 5.1.9.RELEASE, and Reactor version Californium.SR11. That's what I've updated SC OSB to use.

Albertoimpl: We definitely should. Sadly, dependabot was not smart enough to parse that conditional. We should take a look at the other projects too.
Thanks for noticing.

build.gradle
```diff
@@ -20,11 +20,10 @@ buildscript {
 	}
 	dependencies {
 		classpath("io.spring.gradle:propdeps-plugin:0.0.10.RELEASE")
-		classpath("io.spring.gradle:dependency-management-plugin:1.0.5.RELEASE")
-		classpath("io.spring.gradle:spring-io-plugin:0.0.8.RELEASE")
```
royclarkson: Thanks for roving this. we don't need to support it in any project any longer.

build.gradle
```diff
@@ -51,51 +50,21 @@ configure(allprojects) {
 	apply plugin: "io.spring.dependency-management"
 
 	ext {
-		springBootVersion = project.findProperty("springBootVersion") ?: "2.1.4.RELEASE"
-		springVersion = project.findProperty("springVersion") ?: "5.1.6.RELEASE"
-		reactorVersion = project.findProperty("reactorVersion") ?: "Californium-SR6"
-		openServiceBrokerVersion = "3.0.0.RELEASE"
+		springBootVersion = "2.1.8.RELEASE"
```
royclarkson: We should keep these overrides. It's not overriding anything by default, but it's allowing us to override the version in CI or for testing. For example, we have a nightly CI build for spring cloud open service broker that builds and verifies compatibility with the latest boot and framework snapshots.

Albertoimpl: That's true but we are not using them at the moment. I am happy to create a chore to create the concourse CI job for it.
What do you think @royclarkson?

royclarkson: Good idea. We'll need to keep the 1.0.x releases of App Broker on Boot 2.1.x, so maintaining a nightly CI build that verifies everything works on the latest releases or snapshots is pretty valuable.

Albertoimpl: Here it is: https://github.com/spring-cloud/spring-cloud-app-broker/issues/274

PR #279, opened 2019-10-21
Removing request from logging

### LINKED ISSUE #278: Sensitive configuration should not be logged by default
When creating a service broker for Spring Cloud Config Server, we found that when sensitive configuration information caused an error during create/update that it was logged:

```
2019-10-02 16:37:09.110 ERROR 8 --- o.s.c.a.s.WorkflowServiceInstanceService : Error creating service instance ServiceBrokerRequest{platformInstan
ceId='null', apiInfoLocation='api.example.io/v2/info', originatingIdentity=null}AsyncServiceBrokerRequest{asyncAccepted=true}AsyncParameterizedService
InstanceRequest{parameters={git={label=master, uri=git@example.io:acceptance/spring-cloud-config.git, cloneOnStart=true, privateKey=-----BEGIN RSA PRI
VATE KEY-----
MIIEpAIBAAKCAQEAoqyz6YaYMTr7L8GLPSQpAQXaM04gRx4CCsGK2kfLQdw4BlqI...
```

For many customers this could be a problem for their security policy validation rules. App Broker should have a way to ensure configuration with sensitive data is not logged or put anywhere that it might be accessible during or after processing the request.

A first step could be to remove any offending log statements. A secondary step could be to provide a way to turn on/off as a user of Spring Cloud App Broker or enable redacting of specific pieces of configuration although could be error prone for future changes.

Albertoimpl:
[Fixes #278]


Albertoimpl:
Good point! Just pushed it. Thanks!

PR #282, opened 2019-10-28
Bumping assertJ and mockito

Albertoimpl:


Albertoimpl:
We were leaving boot to manage all the modules but in `spring-cloud-app-broker-core`. Which is where the dependencies were listed out.
We have a few options:
- List them everywhere and upgrade them everywhere
- Leaving it to boot everywhere but in `spring-cloud-app-broker-core` and leaving the version managed by boot in `spring-cloud-app-broker-core`, in other words, close this PR and do nothing.
- Leaving it to boot everywhere but in `spring-cloud-app-broker-core` and upgrading only here, meaning, we will have different versions everywhere.

What do you think @royclarkson?

royclarkson:
Unless there is a good reason otherwise, my preference is to keep `spring-cloud-app-broker-core` consistent with the dependencies managed by boot in the other modules. That's how I've been managing those in SCOSB.

Albertoimpl:
That's fair. Thanks for the feedback @royclarkson!

Issue #285, opened 2019-10-30
Interests in orange contributions supporting osb-cmdb use-cases ?


gberche-orange:
Thanks for sharing SCAB with the community. This is an impressive piece of work!

I conducted a spike to use SCAB for Orange's osb-cmdb use-case (see https://github.com/orange-cloudfoundry/paas-templates/issues/492 for background) and SCAB was mostly matching the use-cases with the following few changes (which are visible on the [orange-cloudfoundry/osb-cmdb-spike repo](https://github.com/orange-cloudfoundry/osb-cmdb-spike) see [diff](https://github.com/spring-cloud/spring-cloud-app-broker/compare/master...orange-cloudfoundry:master)):
- support brokered services without backing applications (i.e. only backing services) https://github.com/spring-cloud/spring-cloud-app-broker/pull/288
- support asynchronous backing services (i.e. increase the default 5min cf-java-client timeout to fetch service instance status) #289
- provide ability to AppDeploymentCreateServiceBindingWorkflow to provision service keys (through extending Deployer service or potentially opening up some base classes to enable extensions without duplicating them)   : #292 
- support for brokered service plan upgrades cascading to backing service. See #290

Additionally, there are the following pending issues/use-cases that I would need to handle:
- Fully support async backing services: enable SCAB to provide operation status even in case of SCAB broker restart 
- More flexibility into where backing services are created: See #295
   - Target isn't supporting dynamically created Space yet shared among multiple service instances.
   - Impl is currently hardcoded into `CloudFoundryAppDeployer.createServiceInstance()` by looking up properties (some changes required to support by strategy definitions)
- CreateServiceInstance/UpdateServiceInstance does not yet return dashboard url for brokered services 
  - Need to subclass workflow impls buildResponse() to set dashboard URL.
- Support for dynamically constructed catalog: See #317
  - Need to inject BrokeredServices, see #311

If there is interest in Orange's contributions to SCAB to that the osb-cmdb repo can use SCAB upstream version, I will to submit individual issues for changes/use-cases above or directly PRs for simple changes.

Thanks in advance!

royclarkson:
@gberche-orange thanks for the feedback. Generally speaking, we're happy to discuss any ideas for improving the project. I haven't had a chance to review your suggestions in detail yet.

gberche-orange:
[...]

PR #289, opened 2019-10-31
Configurable API polling duration to support async backing services

gberche-orange:
By default cf-java-client will set a default completion timeout of 5 Mins.

The api-polling-timeout property now enables a configureable value
applied in create/update/delete service instance CF API calls for
backing services.

See https://github.com/spring-cloud/spring-cloud-app-broker/issues/285 for more background on related use-case

### DIFF
--- a/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
+++ b/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
@@ -975,6 +975,7 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 				.serviceName(request.getName())
 				.planName(request.getPlan())
 				.parameters(request.getParameters())
+				.completionTimeout(Duration.ofSeconds(this.defaultDeploymentProperties.getApiPollingTimeout()))
 				.build();
 
 		Mono<CreateServiceInstanceResponse> createServiceInstanceResponseMono =
@@ -1037,6 +1038,7 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 			org.cloudfoundry.operations.services.DeleteServiceInstanceRequest
 				.builder()
 				.name(serviceInstanceName)
+				.completionTimeout(Duration.ofSeconds(this.defaultDeploymentProperties.getApiPollingTimeout()))
 				.build())
 			.doOnError(exception -> LOG.debug(String.format("Error deleting service instance %s with error '%s'",
 				serviceInstanceName, exception.getMessage()), exception))
@@ -1108,6 +1110,7 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 		return cloudFoundryOperations.services().updateInstance(
 			org.cloudfoundry.operations.services.UpdateServiceInstanceRequest.builder()
 				.serviceInstanceName(serviceInstanceName)
+				.completionTimeout(Duration.ofSeconds(this.defaultDeploymentProperties.getApiPollingTimeout()))
 				.parameters(request.getParameters())
 				.build())
 			.then(Mono.just(UpdateServiceInstanceResponse.builder()
--- a/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
+++ b/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
@@ -796,6 +796,13 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 			.orElse(this.defaultDeploymentProperties.getHealthCheckTimeout());
 	}
 
+	private Duration apiPollingTimeout(Map<String, String> properties) {
+		return Duration.ofSeconds(
+			Optional.ofNullable(properties.get(CloudFoundryDeploymentProperties.API_POLLING_TIMEOUT_PROPERTY_KEY))
+			.map(Long::parseLong)
+			.orElse(this.defaultDeploymentProperties.getApiPollingTimeout()));
+	}
+
 	private Integer instances(Map<String, String> properties) {
 		return Optional.ofNullable(properties.get(DeploymentProperties.COUNT_PROPERTY_KEY))
 			.map(Integer::parseInt)
@@ -975,7 +982,7 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 				.serviceName(request.getName())
 				.planName(request.getPlan())
 				.parameters(request.getParameters())
-				.completionTimeout(Duration.ofSeconds(this.defaultDeploymentProperties.getApiPollingTimeout()))
+				.completionTimeout(apiPollingTimeout(request.getProperties()))
 				.build();
 
 		Mono<CreateServiceInstanceResponse> createServiceInstanceResponseMono =
@@ -1016,12 +1023,12 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 			String space = deploymentProperties.get(DeploymentProperties.TARGET_PROPERTY_KEY);
 			requestDeleteServiceInstance = operationsUtils.getOperations(deploymentProperties)
 				.flatMap(cfOperations -> unbindServiceInstance(serviceInstanceName, cfOperations)
-					.then(deleteServiceInstance(serviceInstanceName, cfOperations)))
+					.then(deleteServiceInstance(serviceInstanceName, cfOperations, deploymentProperties)))
 				.then(deleteSpace(space));
 		}
 		else {
 			requestDeleteServiceInstance = unbindServiceInstance(serviceInstanceName, operations)
-				.then(deleteServiceInstance(serviceInstanceName, operations));
+				.then(deleteServiceInstance(serviceInstanceName, operations, deploymentProperties));
 		}
 
 		return requestDeleteServiceInstance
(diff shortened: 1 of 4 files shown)


spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
```diff
@@ -136,6 +138,7 @@
 	@BeforeEach
 	void setUp() {
 		deploymentProperties = new CloudFoundryDeploymentProperties();
+		deploymentProperties.setApiPollingTimeout(EXPECTED_COMPLETION_DURATION); //set a different value than the default to make sure it gets loaded
```
Albertoimpl: Please, remove the comment, the name seems explicit enough.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
```diff
@@ -55,6 +55,8 @@
 
 	static final String JAVA_OPTS_PROPERTY_KEY = "javaOpts";
 
+	public static final long DEFAULT_API_POLLING_TIMEOUT_SECONDS = Duration.ofMinutes(5).getSeconds(); // Default in cf-java-client when no completionTimeout is specified in CSI/USI/DSI
```
Albertoimpl: Please, remove the comment, the name seems explicit enough.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
```diff
@@ -97,10 +99,15 @@
...
 	/**
-	 * Timeout for blocking API calls, in seconds.
+	 * Timeout for blocking CF API calls, in seconds.
 	 */
 	private long apiTimeout = 360L;
 
+	/**
+	 * Timeout for polled async CF API calls, in seconds. Named "completionTimeout" in cf-java-client
+	 */
+	private long apiPollingTimeout = DEFAULT_API_POLLING_TIMEOUT_SECONDS;
```
Albertoimpl: Since we have `api-timeout` which relates to the backing application timeout and it is documented as
```
The timeout value used for blocking CF API calls, in seconds.
```

Adding now `api-polling-timeout` feels like it is related to backing applications when it isn't. Shall we set a name that is more explicit to backing services to make the difference more obvious?
Something in the line of: `backingServiceCompletionTimeout`. What do you think?

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -941,6 +941,7 @@ private Path getApplication(Resource resource) {
 				.serviceName(request.getName())
 				.planName(request.getPlan())
 				.parameters(request.getParameters())
+				.completionTimeout(Duration.ofSeconds(this.defaultDeploymentProperties.getApiPollingTimeout()))
```
Albertoimpl: This code will just set the default to the time specified in, which is great but not what was documented.
If we want a property to be passed through we need to do something like https://github.com/spring-cloud/spring-cloud-app-broker/blob/master/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java#L792-L796 where we check if we have a value that was passed to the `Deployer` layer, otherwise, we use the provided default.

spring-cloud-app-broker-docs/src/docs/asciidoc/service-instances.adoc
```diff
@@ -29,6 +29,7 @@ spring:
             *health-check: http*
             *health-check-http-endpoint: /health*
             *health-check-timeout: 180*
+            *api-polling-timeout: 300*
```
Albertoimpl: See https://github.com/spring-cloud/spring-cloud-app-broker/pull/289/files#r351238763

Albertoimpl:
Thanks a lot for the contributions @gberche-orange!
They are really useful and I would love to merge them.

Left a few comments. I would like to debate with you about this particular one.

Also, do you mind rebasing and applying the formatting we have in https://github.com/spring-cloud/spring-cloud-app-broker/tree/master/src so that we can merge this in when ready?

Thanks a lot!

gberche-orange:
Thanks a lot @Albertoimpl for your review and comments! I hope to have time available to address them  within a week, I'll update the issue.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
gberche-orange: Many other CF java client requests have a configurable completion timeout beyond services, even though they are not net leveraged in SCAB e.g.:
- [DeleteApplicationRequest](https://github.com/cloudfoundry/cf-java-client/blob/bb3de77403d7c79daab360d715cc6113e90c2ce7/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/applications/_DeleteApplicationRequest.java#L34)
- [DeleteSpaceRequest](https://github.com/cloudfoundry/cf-java-client/blob/bb3de77403d7c79daab360d715cc6113e90c2ce7/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/spaces/_DeleteSpaceRequest.java#L33)
- [DeleteRouteRequest](https://github.com/cloudfoundry/cf-java-client/blob/bb3de77403d7c79daab360d715cc6113e90c2ce7/cloudfoundry-operations/src/main/java/org/cloudfoundry/operations/routes/_DeleteRouteRequest.java#L41)

By naming the property generically (and not specifically to services), I was hoping this property could also be applied to other cf-java-client requests in the future

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
gberche-orange: Thanks for catching this. I'm preparing a fix to indeed fetch the configured from properties.

> Also, note that all the properties passed to CloudFoundryDeploymentProperties were always application-related this will be the first and only one related to services.

Please let me know if this would be ok to leave this property generically named in CloudFoundryDeploymentProperties, or what alternative would be recommended instead.

Note that Javadoc seems to assume brokered services could be configured with this property https://github.com/spring-cloud/spring-cloud-app-broker/blob/171554905e2a6eb39d4082be42d131fd33a0a4e7/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java#L28-L29

spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
gberche-orange: thanks, this is fixed with  bd25634

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
gberche-orange: thanks, this is fixed with  bd25634

spring-cloud-app-broker-docs/src/docs/asciidoc/service-instances.adoc
gberche-orange: resolving to avoid duplicate conversations about the property naming

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
Albertoimpl: I see it now. Thanks for explaining!
If we wanted to introduce some of the more specific, we could do something like:
by default use the value of `DEFAULT_API_POLLING_TIMEOUT_SECONDS`
if a value was provided for `api-polling-timeout` use that.
if a value for a specific request was provided, overwrite `api-polling-timeout` with that new property value.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
Albertoimpl: Go with the name you added. Thanks for explaining!

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
gberche-orange: Thanks! Your proposal to use api polling timeout as a default for future more specific properties makes sense. I'm therefore resolving the conversation, please reopen if there is something missing.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
gberche-orange: Thanks.

The fix to fetch the timeout from configured properties is in  8455458
I'm therefore resolving the conversation, please reopen if there is something missing.

gberche-orange:
sorry @Albertoimpl for the delay in allocating time to this PR. 

I hope to have addressed your valuable comments. 

Note that I have intentionally not squashed commits in order to make review of incremental changes easier, and rather I have left that squashing up to the github [PR squash and merge](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) step.

Albertoimpl:
Looking good! Thanks a lot!

gberche-orange:
Thanks for your kind review and merge @Albertoimpl !

PR #290, opened 2019-11-04
Support for brokered service plan update

gberche-orange:
This PR provides ability for a brokered service plan update to cascade the plan update to backing services. See https://github.com/spring-cloud/spring-cloud-app-broker/issues/285 for more background around this use-case.

In the master branch, a brokered service update only triggers a backing service update if params were changed. I understand this is an optimization which avoid extra calls to CF

https://github.com/spring-cloud/spring-cloud-app-broker/blob/c32dd2b103ff691d762af082c30194f1774872b0/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java#L1061-L1066

Initial branch (available at https://github.com/orange-cloudfoundry/osb-cmdb-spike/tree/updateable-plans-not-optimized) discarded the optimization. I started updating the integration tests and failed to update the wiremock stub tests within a reasonable time.

Alternatives considered for preserving the optimization
1. store a previousPlan (String) in BackingService, so that org.springframework.cloud.appbroker.deployer.DeployerClient.updateServiceInstance() can optionally set the plan to non null in org.springframework.cloud.appbroker.deployer.UpdateServiceInstanceRequest
1. fetch backing service instance using [GET /v2/service_instances/:instance_id](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#fetching-a-service-instance) and compare with `UpdateServiceInstanceRequest`. This systematic extra OSB fetch call defeats the optimization.

I then worked to preserve the optimization with approach 1 above

### DIFF
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
@@ -29,11 +29,11 @@ public class BackingService {
 
 	private String serviceInstanceName;
 
-	private String name; // a service definition
+	private String name;
 
 	private String plan;
 
-	private String previousPlan; // Necessary to trigger update on plan mutation.
+	private String previousPlan;
 
 	private Map<String, Object> parameters;
 
@@ -132,10 +132,6 @@ public class BackingService {
 		this.rebindOnUpdate = rebindOnUpdate;
 	}
 
-	/**
-	 * Returns the requested updated plan if any. This enables to not trigger backing service when no plan update was
-	 * requested.
-	 */
 	public String getUpdatedPlanIfAny() {
 		if (getPlan() == null) {
 			return null; //should only happen in incomplete unit tests
@@ -145,7 +141,7 @@ public class BackingService {
 		if (planWasUpdated) {
 			return getPlan();
 		}
-		return null;
+		return null; //Don't trigger backing service update when no plan update was requested.
 	}
 
 	@Override
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
@@ -29,9 +29,11 @@ public class BackingService {
 
 	private String serviceInstanceName;
 
-	private String name;
+	private String name; // a service definition
 
 	private String plan;
+
+	private String previousPlan; // Necessary to trigger update on plan mutation.
 
 	private Map<String, Object> parameters;
 
@@ -47,6 +49,7 @@ public class BackingService {
 	public BackingService(String serviceInstanceName,
 		String name,
 		String plan,
+		String previousPlan,
 		Map<String, Object> parameters,
 		Map<String, String> properties,
 		List<ParametersTransformerSpec> parametersTransformers,
@@ -54,6 +57,7 @@ public class BackingService {
 		this.serviceInstanceName = serviceInstanceName;
 		this.name = name;
 		this.plan = plan;
+		this.previousPlan = previousPlan;
 		this.parameters = parameters;
 		this.properties = properties;
 		this.parametersTransformers = parametersTransformers;
@@ -82,6 +86,14 @@ public class BackingService {
 
 	public void setPlan(String plan) {
 		this.plan = plan;
+	}
+
+	public String getPreviousPlan() {
+		return previousPlan;
+	}
+
+	public void setPreviousPlan(String previousPlan) {
+		this.previousPlan = previousPlan;
 	}
 
 	public Map<String, Object> getParameters() {
@@ -120,6 +132,22 @@ public class BackingService {
 		this.rebindOnUpdate = rebindOnUpdate;
 	}
 
+	/**
+	 * Returns the requested updated plan if any. This enables to not trigger backing service when no plan update was
+	 * requested.
+	 */
+	public String getUpdatedPlanIfAny() {
+		if (getPlan() == null) {
+			return null; //should only happen in incomplete unit tests
+		}
+
+		boolean planWasUpdated = getPreviousPlan() != null && !getPlan().equals(getPreviousPlan());
+		if (planWasUpdated) {
+			return getPlan();
+		}
+		return null;
+	}
+
 	@Override
 	public boolean equals(Object o) {
 		if (this == o) {
@@ -132,6 +160,7 @@ public class BackingService {
 		return Objects.equals(serviceInstanceName, that.serviceInstanceName) &&
 			Objects.equals(name, that.name) &&
 			Objects.equals(plan, that.plan) &&
+			Objects.equals(previousPlan, that.previousPlan) &&
 			Objects.equals(parameters, that.parameters) &&
 			Objects.equals(properties, that.properties) &&
 			Objects.equals(parametersTransformers, that.parametersTransformers) &&
@@ -141,7 +170,8 @@ public class BackingService {
 	@Override
 	public int hashCode() {
 		return Objects
-			.hash(serviceInstanceName, name, plan, parameters, properties, parametersTransformers, rebindOnUpdate);
+			.hash(serviceInstanceName, name, plan, previousPlan, parameters, properties, parametersTransformers,
+				rebindOnUpdate);
 	}
 
 	@Override
@@ -150,6 +180,7 @@ public class BackingService {
 			"serviceInstanceName='" + serviceInstanceName + '\'' +
 			", name='" + name + '\'' +
 			", plan='" + plan + '\'' +
+			", previousPlan='" + previousPlan + '\'' +
 			", parameters=" + parameters +
 			", properties=" + properties +
 			", parametersTransformers=" + parametersTransformers +
@@ -169,6 +200,8 @@ public class BackingService {
 
 		private String plan;
 
+		private String previousPlan;
+
 		private final Map<String, Object> parameters = new HashMap<>();
 
 		private final Map<String, String> properties = new HashMap<>();
@@ -184,6 +217,7 @@ public class BackingService {
 			return this.serviceInstanceName(backingService.getServiceInstanceName())
 				.name(backingService.getName())
 				.plan(backingService.getPlan())
+				.previousPlan(backingService.getPreviousPlan())
 				.parameters(backingService.getParameters())
 				.properties(backingService.getProperties())
 				.parameterTransformers(backingService.getParametersTransformers())
@@ -202,6 +236,11 @@ public class BackingService {
 
 		public BackingServiceBuilder plan(String plan) {
 			this.plan = plan;
+			return this;
+		}
+
+		public BackingServiceBuilder previousPlan(String previousPlan) {
+			this.previousPlan = previousPlan;
 			return this;
 		}
 
@@ -239,7 +278,8 @@ public class BackingService {
 		}
 
 		public BackingService build() {
-			return new BackingService(serviceInstanceName, name, plan, parameters, properties, parameterTransformers,
+			return new BackingService(serviceInstanceName, name, plan, previousPlan, parameters, properties,
+				parameterTransformers,
 				rebindOnUpdate);
 		}
 
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
@@ -16,22 +16,28 @@
 
 package org.springframework.cloud.appbroker.workflow.instance;
 
+import java.util.List;
+
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.util.Logger;
 import reactor.util.Loggers;
 
 import org.springframework.cloud.appbroker.deployer.BackingAppDeploymentService;
+import org.springframework.cloud.appbroker.deployer.BackingService;
 import org.springframework.cloud.appbroker.deployer.BackingServicesProvisionService;
 import org.springframework.cloud.appbroker.deployer.BrokeredServices;
 import org.springframework.cloud.appbroker.extensions.parameters.BackingApplicationsParametersTransformationService;
 import org.springframework.cloud.appbroker.extensions.parameters.BackingServicesParametersTransformationService;
 import org.springframework.cloud.appbroker.extensions.targets.TargetService;
 import org.springframework.cloud.appbroker.service.UpdateServiceInstanceWorkflow;
+import org.springframework.cloud.servicebroker.model.catalog.Plan;
+import org.springframework.cloud.servicebroker.model.catalog.ServiceDefinition;
 import org.springframework.cloud.servicebroker.model.instance.UpdateServiceInstanceRequest;
 import org.springframework.cloud.servicebroker.model.instance.UpdateServiceInstanceResponse;
 import org.springframework.cloud.servicebroker.model.instance.UpdateServiceInstanceResponse.UpdateServiceInstanceResponseBuilder;
 import org.springframework.core.annotation.Order;
+
 
 @Order(0)
 public class AppDeploymentUpdateServiceInstanceWorkflow extends AppDeploymentInstanceWorkflow
@@ -49,18 +55,22 @@ public class AppDeploymentUpdateServiceInstanceWorkflow extends AppDeploymentIns
 
 	private final TargetService targetService;
 
+	private final BackingServicesUpdateValidatorService servicesUpdateValidatorService;
+
 	public AppDeploymentUpdateServiceInstanceWorkflow(BrokeredServices brokeredServices,
 		BackingAppDeploymentService deploymentService,
 		BackingServicesProvisionService backingServicesProvisionService,
 		BackingApplicationsParametersTransformationService appsParametersTransformationService,
 		BackingServicesParametersTransformationService servicesParametersTransformationService,
-		TargetService targetService) {
+		TargetService targetService,
+		BackingServicesUpdateValidatorService backingServicesUpdateValidatorService) {
 		super(brokeredServices);
 		this.deploymentService = deploymentService;
 		this.backingServicesProvisionService = backingServicesProvisionService;
 		this.appsParametersTransformationService = appsParametersTransformationService;
 		this.servicesParametersTransformationService = servicesParametersTransformationService;
 		this.targetService = targetService;
+		servicesUpdateValidatorService = backingServicesUpdateValidatorService;
 	}
 
 	@Override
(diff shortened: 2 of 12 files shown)


gberche-orange:
I started implementation of approach 1 above. The plan upgrade logic would check that a requested plan update to a brokered service only translates into a backing service plan update. In other words, brokered service plan update would be rejected if resulting brokered services differs in name or service definition, i.e. 
* new backing service creation
* deleted backing service 

Pseudo code would be:
- previousBackingServices = list of backing services with previous plan;
- candidateBackingServicesForUpdate = list of backing services with requested plan
- Reject if number of elements differ in the two list
- Iterate over one list and reject if there is not exactly matching element with same instance name and service definition

Unit testing this logic in the AppDeploymentUpdateServiceInstanceWorkflow is time consumming though because:
- There are 6 collaborators with pretty strict mocked interactions. Required modifying mocks set up and assertion to introduce new feature
- Uses private methods (updateBackingServices and updateBackingApplications) which can't be tested individually
- AppDeploymentUpdateServiceInstanceWorkflow extends AppDeploymentInstanceWorkflow which provides private methods (such as `getBackingServicesForService()` or `getBackingApplicationsForService()` which can't be used in another service

Considered alternatives:
1. Don't unit test the plan upgrade validation logic, and keep it in AppDeploymentUpdateServiceInstanceWorkflow with only high level testing
1. Unit test the plan upgrade validation logic by making the suppporting method visible to unit test (protected or package protected visibility)
1. Introduce a new `BackingServicesUpdateValidatorService` to unit test the plan update logic
   1. Refactor AppDeploymentInstanceWorkflow to extract logic into a BrokeredServicesFinderService bean
   1. Duplicate brokered service helpers methods from AppDeploymentInstanceWorkflow 
   1. Modify AppDeploymentInstanceWorkflow  to compute candidateBackingServicesForUpdate  and previousBackingServices 

Finally selected option 3iii.

gberche-orange:
Tests are now passing :-)

Here are improvements that this PR could use with more available time
* Refine user-facing log messages when plan updates get rejected
* Add tests for more edge cases (e.g. an UpdateServiceInstanceRequest with a previousPlan not resolving to any known backing service)
* Simplify implementation of BackingServicesUpdateValidatorService with more idomatic functional styling
* Change BackingService.previousPlan into a boolean flag such as  BackingService.planWasUpdated
* Refactor AppDeploymentUpdateServiceInstanceWorkflow.updateBackingServices() to make it smaller
* Simplify the unit tests (DeployerClientTest, AppDeploymentUpdateServiceInstanceWorkflowTest, CloudFoundryAppDeployerTest) to use less mocks or have more flexible stubbing/mocking assertions. 
* Refactor 

When trying to implement this feature, the mocked collaborators were often making the test fragile, causing hard-to-diagnose false negatives and making the introduction of new behaviors harder. This feels like popular related discussions such as https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html 

The tests above could probably use a Fake or a simple impl for TargetService, *ParametersTransformationService instead of mocks.

Albertoimpl:
Thanks for the contribution, would you mind rebasing this PR @gberche-orange? Thanks!

gberche-orange:
@Albertoimpl the PR was rebased against master and adapted to new checkstyles and pmd rules. I'm looking forward to your review.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
```diff
@@ -0,0 +1,79 @@
...
+		try {
+			return Mono.just(validateAndMutatePlanUpdate(previousBackingServices.block(),
+				candidateBackingServicesForUpdate.block()));
+		}
+		catch (IllegalArgumentException e) {
+			return Mono.error(e);
+		}
+	}
+
+	public List<BackingService> validateAndMutatePlanUpdate(List<BackingService> previousBackingServicesList,
```
Albertoimpl: This should be private and the tests should call the parent method.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.cloud.appbroker.deployer.BackingService;
+
+/**
+ * Validates updates to backing services, and returns an updated copy of the list with the previousPlan set. This help
+ * only triggering plan update when previousPlan is set.
+ */
+public class BackingServicesUpdateValidatorService {
+
+	// Accept and return Mono to make it easier for AppDeploymentUpdateServiceInstanceWorkflow to provide us with previous and candidate backing services
+	public Mono<List<BackingService>> validatePlanUpdate(Mono<List<BackingService>> previousBackingServices,
```
Albertoimpl: `validatePlanUpdate` is not really just validating, actually, it doesn't really validate that the new parameters passed through the new `candidateBackingServicesForUpdate` are valid at all.

I would say this class is filtering the services to update, which is great, does one thing and does it well. Just the naming is what I find misleading.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.springframework.cloud.appbroker.deployer.BackingService;
+
+/**
+ * Validates updates to backing services, and returns an updated copy of the list with the previousPlan set. This help
+ * only triggering plan update when previousPlan is set.
+ */
+public class BackingServicesUpdateValidatorService {
+
+	// Accept and return Mono to make it easier for AppDeploymentUpdateServiceInstanceWorkflow to provide us with previous and candidate backing services
```
Albertoimpl: Remove the comment, please

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	private BackingService getValidatedBackingService(BackingService candidate,
+		List<BackingService> previousBackingServicesList) {
+		List<BackingService> matchingPreviousBackingServices = previousBackingServicesList.stream()
+			.filter(previousBackingService ->
+				previousBackingService.getName().equals(candidate.getName()) &&
+					previousBackingService.getServiceInstanceName().equals(candidate.getServiceInstanceName()))
+			.collect(Collectors.toList());
+		if (matchingPreviousBackingServices.size() != 1) { // NOPMD
```
Albertoimpl: Any reason to disallow this completely?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
```diff
@@ -40,6 +40,10 @@
 		return Mono.empty();
 	}
 
+	/**
```
Albertoimpl: Please, delete the comment to match the other methods.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
```diff
@@ -120,6 +132,22 @@ public void setRebindOnUpdate(boolean rebindOnUpdate) {
 		this.rebindOnUpdate = rebindOnUpdate;
 	}
 
+	/**
+	 * Returns the requested updated plan if any. This enables to not trigger backing service when no plan update was
+	 * requested.
+	 */
+	public String getUpdatedPlanIfAny() {
+		if (getPlan() == null) {
```
Albertoimpl: Can we update the tests instead of adding a branch in the code that will never be hit?

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
```diff
@@ -29,10 +29,12 @@
 
 	private String serviceInstanceName;
 
-	private String name;
+	private String name; // a service definition
```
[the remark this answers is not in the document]
gberche-orange: I felt the field `name` was ambiguous and would benefit from clarification. I'll rather submit an independent PR just for renaming this field into serviceDefinitionName as to keep this PR focused and small.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
gberche-orange: The reason to reject such update is that other types of upgrades (reported in #293 and illustrated into `BackingServicesUpdateValidatorServiceTest` ) would probably not be properly fullfilled by the CloudFoundryAppDeployer such as:
* deleting removed backing services
* changing type of backing service (e.g. from mysql to postgresql)
* changing service instance name (i.e. renaming a service instance)

May be the case of adding new backing services could be supported in the future with additional work.

To keep it simple as a first step, I was favoring a "fail-fast" approach.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
gberche-orange: Thanks. I fell this would be duplicating a lot of wrapping in/out of Mono in unit tests. I've rather inlined the validatePlanUpdate() method to caller and only keeping validateAndMutatePlanUpdate() as a single public method.

Let me know if you find a better alternative.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
```diff
@@ -120,6 +132,22 @@ public void setRebindOnUpdate(boolean rebindOnUpdate) {
 		this.rebindOnUpdate = rebindOnUpdate;
 	}
 
+	/**
+	 * Returns the requested updated plan if any. This enables to not trigger backing service when no plan update was
```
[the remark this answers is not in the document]
gberche-orange: I felt the design rationale for having this field (i.e. to avoid triggering unnecessary update) would not be obvious to other readers of the codebase. 

I've therefore moved this comment to the "return null" statement, as to preserve consistency in lack of javadoc in this class.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
```diff
@@ -70,15 +80,50 @@ public AppDeploymentUpdateServiceInstanceWorkflow(BrokeredServices brokeredServi
 			.then();
 	}
 
+	//Inspired from  spring-cloud-open-service-broker private method
```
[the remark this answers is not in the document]
gberche-orange: thanks

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
gberche-orange: The actual validation is performed in the `getValidatedBackingService()` which will throw an exception, triggering the emission of an error in the flux.

Also, since the provided BackingService bean has not a request scope (i.e. is reused among requests), it returns a mutated copy with the updated plan set.

Alternative namings I can think of:
* `BackingServicesUpdateFactoryService`: expresses that a new `BackingService` list is created, but misses the fact that an exception is thrown when the user request is invalid/not supported
   * `validateAndMutatePlanUpdate()`could become `validateAndCreateUpdatedBackingServices()`

Please let me know if you find a better alternative naming and I'll then apply the change once (as to avoid back and forth changes)

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
gberche-orange: thanks

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
gberche-orange: I was trying to make it explicit the non-obvious behavior that the mono might be empty when the update is optimized and skipped, which is currently only visible in the implementation by reading https://github.com/spring-cloud/spring-cloud-app-broker/blob/76765d10775856930874badf430f70e8c3967614/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java#L1107-L1114
To me this breaks encapsulation between the modules.

What could be a better alternative to this seldom comment ?

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
```diff
@@ -29,10 +29,12 @@
 
 	private String serviceInstanceName;
 
-	private String name;
+	private String name; // a service definition
 
 	private String plan;
 
+	private String previousPlan; // Necessary to trigger update on plan mutation.
```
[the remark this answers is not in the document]
gberche-orange: thanks

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
```diff
@@ -0,0 +1,79 @@
...
+ * Validates updates to backing services, and returns an updated copy of the list with the previousPlan set. This help
+ * only triggering plan update when previousPlan is set.
+ */
+public class BackingServicesUpdateValidatorService {
+
+	// Accept and return Mono to make it easier for AppDeploymentUpdateServiceInstanceWorkflow to provide us with previous and candidate backing services
+	public Mono<List<BackingService>> validatePlanUpdate(Mono<List<BackingService>> previousBackingServices,
+		Mono<List<BackingService>> candidateBackingServicesForUpdate) {
+		try {
+			return Mono.just(validateAndMutatePlanUpdate(previousBackingServices.block(),
```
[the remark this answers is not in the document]
gberche-orange: thanks!

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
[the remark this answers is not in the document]
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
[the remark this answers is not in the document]
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/BackingServicesUpdateValidatorService.java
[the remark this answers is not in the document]
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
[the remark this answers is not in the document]
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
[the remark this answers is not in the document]
gberche-orange: fixed in 661f7ed3e6ef425dfdd7baabd18f96f6b4d7c287

gberche-orange:
Thanks @Albertoimpl for your review.

I've made some cosmetic changes to this PR. I realize however that preserving the update optimization (i.e. Leveraging the previousValues from the update service instance request) has the following impacts, and would not trigger expected updates when only metadata changes such as 
   - [maintenance_info](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-6) for triggering security patches
   - [context](https://github.com/openservicebrokerapi/servicebroker/blob/master/profile.md#cloud-foundry-context-object) (such as space/org/instance renaming, or starting in 2.16 changes in annotations)

These updates as important to the osb-cmdb use-cases and likely also for some other SCAB use-cases.

The filtering of updates currently performed in https://github.com/spring-cloud/spring-cloud-app-broker/blob/76765d10775856930874badf430f70e8c3967614/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java#L1107-L1114 could be removed.

If this is valueable to SCAB, I could try again to update the integration tests (which I failed as mentionned in https://github.com/spring-cloud/spring-cloud-app-broker/pull/290#issue-336503901). This will likely require to simplify assertions made against the mocked objects which were preventing any change in behavior as reported in https://github.com/spring-cloud/spring-cloud-app-broker/pull/290#issuecomment-551311140  I may need some help/guideance there.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BackingService.java
gberche-orange: Suspended until discussion in https://github.com/spring-cloud/spring-cloud-app-broker/pull/290#issuecomment-567106156 to remove the filtering CloudFoundryAppDeployer, and therefore this field would be removed

gberche-orange:
My osb-cmdb project is not requiring this app-broker use-case, and I unfortunately I won't be able to maintain the PR to get it merged in the short term. I'll reopen in the future, if I manage to get priority in my other projects to contribute this feature.

Thanks @Albertoimpl for your past support!

PR #291, opened 2019-11-07
Enforce Spring coding style

royclarkson:
- Refactor build.gradle to apply PMD and Checkstyle rules to specific projects
- Add PMD checks and resolve failing rules
- Apply Spring Checkstyle rules based on Spring Framework
- Rearrange imports to be more consistent
- Format whitespace


gberche-orange:
@royclarkson is there an automatic way of applying the same codestyle to pending PRs ? I'm concerned that pending PRs rebase would require extensive conflict resolution.

royclarkson:
@gberche-orange unfortunately this will require some manual cleanup once it's merged. I was expecting to rebase your PRs against the changes, but if you're willing to help with that, then it will go more smoothly. I'm trying to get this PR in today.

gberche-orange:
Thanks @royclarkson 
I have so far rebased and pushed #292. I'm waiting for test feedback. I won't be able to help more today for other branches.

gberche-orange:
@royclarkson is there a known/shareable intellij configuration that can be used to check the PMD and Checkstyle rules and avoid conflicting hints from intellij builtin inspections ? 

Currently, running pmd through gradle make feedback cycles slow, and currently require me to go back and forth between HTML reports and the IDE without quick navigation nor suggested fixes.

image
image


I tried https://plugins.jetbrains.com/plugin/1137-pmdplugin, with importing custom rulesets from ci/config/pmdRuleSet.xml and ci/config/pmdTestRuleSet.xml. This is better but I did not find a way to have it run on all gradle projects, and still manually select the custom rule to run on a set of files

royclarkson:
@gberche-orange I'll update the PR with my IntelliJ settings, but they may not the be the "final" version. There seem to be a few different versions floating around in various spring projects. I agree it's painful. I just went through several days of that for this PR.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/BrokerPropertiesParameterResolver.java
```diff
@@ -18,20 +18,25 @@
...
 import java.lang.reflect.Method;
 import java.util.Optional;
+
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
 
-class BrokerPropertiesParameterResolver implements ParameterResolver {
+public class BrokerPropertiesParameterResolver implements ParameterResolver {
```
Albertoimpl: Any particular reason for making the visibility public over here?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/AppManagementStopAcceptanceTest.java
```diff
@@ -50,25 +53,25 @@ protected String backingServiceName() {
...
 		StepVerifier.create(cloudFoundryService.getServiceInstance(SI_NAME))
-					.assertNext(serviceInstance -> assertThat(serviceInstance.getStatus()).isEqualTo("succeeded"))
-					.verifyComplete();
+			.assertNext(serviceInstance -> assertThat(serviceInstance.getStatus()).isEqualTo("succeeded"))
+			.verifyComplete();
 	}
 
 	@AfterEach
-	void cleanUp() {
+	public void cleanUp() {
```
Albertoimpl: Any particular reason for making the visibility public over here?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/BrokerPropertiesParameterResolver.java
royclarkson: trying to avoid package private

Albertoimpl:
Thanks a lot for all the changes @royclarkson! This is going to make things much easier from now on.

I found a few things that were not added to the `src/idea/spring.xml` file that are required in IntelliJ in order to get the same formatting we have in the PR.

[idea-patch.txt](https://github.com/spring-cloud/spring-cloud-app-broker/files/3830851/idea-patch.txt)

gberche-orange:
@royclarkson In order to not negatively impact contribution velocity, I wonder whether experience and work invested by other spring projects in automated formatting and code style could be leveraged in spring-cloud-app-broker. 

Quickly searching across some other spring projects I ran into https://github.com/spring-io/spring-javaformat that provides 
> A source formatter that applies wrapping and whitespace conventions
> A checkstyle plugin that enforces consistency across a codebase

I wonder whether this was considered and could address some of standard code formatting needs, and be applicable automatically to existing PRs without manual work, and future PRs with IDE support.

Regarding contributors experience for PMD violations, it seems the gradle integration https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.PmdExtension.html supports a console output which could enable viewing PMD violations within the IDE or within the CI test results. Ideally all of PMD violations would be reported at once, instead of failing fast on the first offending gradle subproject. I'm not sure it is possible to interactively run the pmdMain and pmdTest gradle targets and override the `ignoreFailures` to allow the build to continue if there are warnings in one subproject.

There seems some overlap between pmd rules and checkstyle rules, for instance regarding imports:

https://github.com/spring-cloud/spring-cloud-app-broker/blob/5ed5fdaacb4c11763fc6d93318d838e0220ce9cb/src/pmd/pmdRuleSet.xml#L34
https://github.com/spring-cloud/spring-cloud-app-broker/blob/5ed5fdaacb4c11763fc6d93318d838e0220ce9cb/src/pmd/pmdRuleSet.xml#L41

https://github.com/spring-cloud/spring-cloud-app-broker/blob/5ed5fdaacb4c11763fc6d93318d838e0220ce9cb/src/checkstyle/checkstyle.xml#L91-L102

Would it make sense to keep PMD check focused on few seldom areas that checkstyle (possibly through `spring-javaformat` does not cover) ?

royclarkson:
@gberche-orange thanks for the valuable feedback. The last time I reviewed the spring-javaformat plugin, it enforced some formatting that I didn't like. Unfortunately, it's not configurable, so the only choices are use it or not. I'll review it again and compare to what we're trying to do here. Maybe it's the best option anyway, since it offers auto-formatting.

My goal is not to put up roadblocks to contributions, but rather offer consistent expectations about formatting. The check style rules were all imported from Spring Framework as-is (minus the javaformat rules). There is some redundancy with pmd. I'm ok with that as long as it's not causing unnecessary frustration.

royclarkson:
Regarding your last comment, you can add the `--continue` flag to the Gradle build to not fail on the first project, but complete the entire build. It doesn't collect the results into a single report though.

royclarkson:
I applied the spring-javaformat plugin to the SCOSB project as a test, and it is still messing up all the builder/reactive code. It reformats everything to one line, which is far less readable IMHO. See https://github.com/spring-io/spring-javaformat/issues/124. The workaround is to disable the formatter in every place you want to prevent the plugin from reformatting. This isn't ideal for OSB or app broker because each project has a lot of reactive and builder style code. Until this situation improves, I'd prefer to work around it with checkstyle and pmd rules.

gberche-orange:
thanks @royclarkson for your efforts into preserving contributions productivity !

I've commented the spring-javaformat to try to make the situation improve. 

I'll try the IDEA option provided in https://github.com/spring-cloud/spring-cloud-app-broker/blob/spring-style/src/idea/spring.xml presumably following similar procedure than the one documented at https://github.com/spring-cloud/spring-cloud-build#intellij-idea

royclarkson:
@gberche-orange we're more than happy to discuss the addition or exclusion of pmd rules. I already disabled a number that didn't seem to make sense. They're guidelines, not laws. :) Please raise any for review.

PR #297, opened 2019-11-14
Revert "Fix blocking calls to workflow layers"

Albertoimpl:
After adding this change. A lot of our tests started becoming flaking in environments with low resources. Such as in some pipelines. If the tests get executed in parallel in a machine with less memory, some of the tests start to fail:
```
FAILED test: org.springframework.cloud.appbroker.service.WorkflowServiceInstanceBindingServiceTest > createServiceInstanceRouteBindingWithAsyncError()
FAILED test: org.springframework.cloud.appbroker.service.WorkflowServiceInstanceBindingServiceTest > deleteServiceInstanceBinding()
WorkflowServiceInstanceServiceTest. updateServiceInstance()
```
Because of that reason, we want to revert back until we can explain why is this failing. It could be just a matter of the way the code is tested and not the actual code.

This reverts commit 9072f858361be9a2acc8edae659ad98b309a801c.


Albertoimpl:
Before this change. I was able to get only 3/10 successful runs in a resource-constrained environment by running:
```
./gradlew --parallel clean check -x :spring-cloud-app-broker-integration-tests:check
```
With this PR reverting it, it passes 10/10.

Anything against reverting this and give it a second thought on the way we are testing the changes? @GFriedrich @royclarkson

PR #306, opened 2019-12-05
Bumping CFJC to the latest

Albertoimpl:


royclarkson:
This is replaced by #307, which goes to a newer version, so no longer needed.

PR #312, opened 2019-12-12
TargetSpec equals and hashcode

gberche-orange:
This PR provides an implementation for TargetSpec equals() and hashCode() as provided by intellij java7 template.

This supports the #285 and the dynamic catalog, to be able to compare BrokeredService which invoke TargetSpec.equals

https://github.com/spring-cloud/spring-cloud-app-broker/blob/b0c71e52de361ef097c3a6e27e9d63b291a6ed4c/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/deployer/BrokeredService.java#L106


Albertoimpl:
Good catch! Thanks for fixing it.

PR #318, opened 2019-12-20
`UpdateServiceInstance` creates new backing services and/or deletes old backing services

LittleBaiBai:
Still have to fix the stubs for the integration test and add AT.


spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
```diff
@@ -32,6 +32,10 @@
 		return Mono.empty();
 	}
 
+	default Mono<GetApplicationResponse> get(GetApplicationRequest request) {
```
royclarkson: I think maybe `getApplication` is better.

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.util.CollectionUtils;
+
+public class GetApplicationRequest {
+
+	private final String name;
+
+	private final Map<String, String> properties;
```
royclarkson: Should this be a `Map<String, Object>`?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationResponse.java
```diff
@@ -0,0 +1,120 @@
...
+
+	public List<String> getServices() {
+		return services;
+	}
+
+	public static GetDeployedAppResponseBuilder builder() {
+		return new GetDeployedAppResponseBuilder();
+	}
+
+	public static final class GetDeployedAppResponseBuilder {
```
royclarkson: name is inconsistent with class name

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
LittleBaiBai: I initially have the name as `getApplication` as you suggested, but changed it afterwards becaulse all other application related methods in the context don't have `application` in the names, e.g `deploy`, `update`. I agree that `getApplication` is better, should we append `application` for all of them?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
LittleBaiBai: We are using `Map<String, String>` in other files when referring to properties, e.g. `DeployApplicationRequest`, `GetServiceInstanceRequest`. We should keep it consistent.

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationResponse.java
LittleBaiBai: Good catch! Will change it.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -1010,7 +1062,7 @@ private Path getApplication(Resource resource) {
 	public Mono<UpdateServiceInstanceResponse> updateServiceInstance(UpdateServiceInstanceRequest request) {
 		return operationsUtils.getOperations(request.getProperties())
 			.flatMap(cfOperations -> rebindServiceInstanceIfNecessary(request, cfOperations)
-				.then(updateServiceInstanceIfNecessary(request, cfOperations)));
+				.then(updateServiceInstanceIfNecessary(request, cfOperations))); // TODO: is the order right?
```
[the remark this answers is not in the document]
LittleBaiBai: Oh I forgot about that comment. This PR has already grown too big. Would it ok for me to handle this `TODO` in a separate PR?

spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/services/ConditionalBackingAppUpdate.java
```diff
@@ -0,0 +1,50 @@
...
+	private static final String SWAP_BACKING_SERVICE_PARAMETER_KEY = "swap-backing-service";
+
+	private static final Logger LOG = LoggerFactory.getLogger(ConditionalBackingAppUpdate.class);
+
+	@Override
+	public ParametersTransformer<BackingApplication> create(Object config) {
+		return (backingApplication, parameters) -> {
+			if (parameters.containsKey(SWAP_BACKING_SERVICE_PARAMETER_KEY)) {
+				String newServiceName = (String) parameters.get(SWAP_BACKING_SERVICE_PARAMETER_KEY);
+				LOG.info("Going to change backing service to {}", newServiceName);
```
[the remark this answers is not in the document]
LittleBaiBai: 👍

spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/services/ConditionalBackingAppUpdate.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.cloud.appbroker.extensions.parameters.ParametersTransformer;
+import org.springframework.cloud.appbroker.extensions.parameters.ParametersTransformerFactory;
+
+/**
+ * A {@link ParametersTransformerFactory} that adds a new backing service to backing app during update conditionally.
+ */
+public class ConditionalBackingAppUpdate extends ParametersTransformerFactory<BackingApplication, Object> {
+
+	// Importing this value will lead to pmdMain task error
+	private static final String SWAP_BACKING_SERVICE_PARAMETER_KEY = "swap-backing-service";
```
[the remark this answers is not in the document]
LittleBaiBai: I like `replacement-backing-service`. Will update in the next commit.

spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/AppBrokerApplication.java
```diff
@@ -74,13 +77,23 @@ public DeleteServiceInstanceWorkflow deleteServiceInstanceWorkflow() {
...
-	 * A no-op ServiceInstanceBindingService bean
+	 * A parameters transformer bean that conditionally swap out backing service
 	 *
 	 * @return the bean
 	 */
 	@Bean
-	public ServiceInstanceBindingService serviceInstanceBindingService() {
-		return new NoOpServiceInstanceBindingService();
+	public ParametersTransformerFactory<BackingService, ?> conditionalNewBackingService() {
+		return new ConditionalNewBackingService();
```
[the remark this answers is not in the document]
LittleBaiBai: Could you please clarify on what you mean by `model` here?

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
```diff
@@ -79,14 +94,67 @@ public AppDeploymentUpdateServiceInstanceWorkflow(BrokeredServices brokeredServi
...
+			.zipWith(backingAppManagementService.getDeployedBackingApplications(request.getServiceInstanceId())
+				.flatMapMany(Flux::fromIterable)
+				.map(BackingApplication::getServices)
+				.flatMap(Flux::fromIterable)
+				.distinct()
+				.map(servicesSpec -> BackingService.builder()
+					.serviceInstanceName(servicesSpec.getServiceInstanceName())
+					.build())
+				.collectMap(BackingService::getServiceInstanceName, Function.identity()))
+			.flatMapMany(newAndExisting -> {
```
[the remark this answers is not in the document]
LittleBaiBai: I'd love to refactor it as part of this PR. Are there specific changes that you would like to see here?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceWithServicesAcceptanceTest.java
```diff
@@ -90,11 +90,16 @@ void deployAppsAndCreateServicesOnCreateService() {
...
-		// service has no applications bound to it
-		ServiceInstance serviceInstance2AfterDeletion = getServiceInstance(BACKING_SI_2_NAME);
-		assertThat(serviceInstance2AfterDeletion.getApplications()).isEmpty();
-
-		deleteServiceInstance(BACKING_SI_2_NAME);
+		// and the backing services are deleted
+		assertThat(listServiceInstances()).doesNotContain(BACKING_SI_1_NAME);
+		assertThat(listServiceInstances()).doesNotContain(BACKING_SI_2_NAME);
+
+		// TODO: another story to only remove the instances with service definition specified
```
Albertoimpl: Could you add a link to the story as part of the TODO?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/fixtures/cf/CloudFoundryService.java
```diff
@@ -203,6 +204,10 @@ public CloudFoundryService(CloudFoundryClient cloudFoundryClient,
 			.doOnError(error -> LOGGER.error("Error updating service instance " + serviceInstanceName + ": " + error));
 	}
 
+	public Flux<ServiceInstanceSummary> listServiceInstances() {
```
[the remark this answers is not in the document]
Albertoimpl: I see the point of choosing that return type and that name but, as Roy pointed out, I would rather be consistent with what we have.

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/UpdateInstanceWithServicesParametersComponentTest.java
```diff
@@ -33,29 +33,37 @@
...
 
+	protected static final String SERVICE_NAME = "example";
+
+	protected static final String PLAN_NAME = "standard";
+
 	protected static final String BACKING_SI_NAME = "my-db-service";
 
 	protected static final String BACKING_SERVICE_NAME = "db-service";
 
+	protected static final String BACKING_PLAN_NAME = "backing-standard";
```
Albertoimpl: was this file formatted using https://github.com/spring-cloud/spring-cloud-app-broker/blob/master/src/idea/spring.xml ?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
Albertoimpl: Agreed with @LittleBaiBai, we have properties as `Map<String, String>` since they are typically key-value and we have parameters as `Map<String, Object>` since they can be nested.

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2002-2019 the original author or authors
```
Albertoimpl: Let's live in the present 😸.
Probably applies to all the license files. If you rebase master you should get a proper error because of the year.

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
Albertoimpl: I would go with `get`, although I see @royclarkson's point I would rather be consistent and not cause a breaking change.

spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/AppBrokerApplication.java
```diff
@@ -74,13 +77,23 @@ public DeleteServiceInstanceWorkflow deleteServiceInstanceWorkflow() {
...
+	}
+
+	/**
+	 * A parameters transformer bean that conditionally swap out backing service
+	 *
+	 * @return the bean
+	 */
+	@Bean
+	public ParametersTransformerFactory<BackingApplication, ?> conditionalBackingAppUpdate() {
+		return new ConditionalBackingAppUpdate();
```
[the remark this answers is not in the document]
Albertoimpl: I am wondering a couple of things:
- Can this be part of a create or update workflow so that the example is more realistic?
- Can this be part of the test instead of available for all the tests so that it's scoped only for those tests?

Albertoimpl:
@LittleBaiBai  Since this change updates the behaviour we had, would be great to document it as part of this PR.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/UpdateInstanceWithNewServiceAcceptanceTest.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	@Test
+	@AppBrokerTestProperties({
+		"spring.cloud.appbroker.services[0].service-name=" + APP_SERVICE_NAME,
+		"spring.cloud.appbroker.services[0].plan-name=" + PLAN_NAME,
+
+		"spring.cloud.appbroker.services[0].apps[0].name=" + APP_NAME,
+		"spring.cloud.appbroker.services[0].apps[0].path=" + BACKING_APP_PATH,
+		"spring.cloud.appbroker.services[0].apps[0].parameters-transformers[0].name=ConditionalBackingAppUpdate",
```
Albertoimpl: Excellent test @LittleBaiBai! 
But wondering, as mentioned above, if instead of having a parameter-transformer to do the service part wouldn't it be better if we had a workflow as part of this class.

spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/AppBrokerApplication.java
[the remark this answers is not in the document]
Albertoimpl: Iterating over my thoughts, we need a previous state and then apply this new state on an update. So even a workflow might be confusing, wondering if there is something else we can do.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceWithServicesAcceptanceTest.java
LittleBaiBai: Good idea! I will create a story and update the TODO.

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/UpdateInstanceWithServicesParametersComponentTest.java
LittleBaiBai: Good catch! I didn't hit the format key in this file I guess. I tried it just now and added an empty line after this line. Was that what you were referring to?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
LittleBaiBai: Thanks! Moving to present 🐈

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/UpdateInstanceWithServicesParametersComponentTest.java
Albertoimpl: Yep! I just saw that one and got me wondering. You should be able to format the whole projects without any problems if the file was imported anyways.
Thanks!

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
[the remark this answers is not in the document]
LittleBaiBai: New story created here: https://github.com/spring-cloud/spring-cloud-app-broker/issues/320. Removing this `TODO`.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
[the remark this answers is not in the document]
LittleBaiBai: @royclarkson I refactored this class a bit. Does it look better now?

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/GetApplicationRequest.java
royclarkson: Understood. I like consistency. Maybe this should be something we reconsider in 2.x? `String` is still limiting for storing numeric values (unless you store them as strings), regardless of needing to nest properties.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentUpdateServiceInstanceWorkflow.java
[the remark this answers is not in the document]
royclarkson: that looks much more reactive! :)

spring-cloud-app-broker-deployer/src/main/java/org/springframework/cloud/appbroker/deployer/AppDeployer.java
royclarkson: yep, agreed. I wasn't looking at the whole file, so missed the other methods are the same. Thanks.

PR #321, opened 2020-01-27
Truncate app name if too long when using `ServiceInstanceGuidSuffix`

### LINKED ISSUE #204: ServiceInstanceGuidSuffix only works with short service names
It turns out that CF only supports service names up to a length of 50 characters. With a 36 char guid forming part of the final name, that makes it quite likely that users will specify a service name prefix that results in the final name overrunning the allowed number of characters, which results in an error like this:

```
2019-02-13 05:54:07.744 ERROR 8 --- [ry-client-nio-2] pDeploymentCreateServiceInstanceWorkflow : Error creating backing services for p.config-server-alpha/standard with error 'CF-ServiceInstanceNameTooLong(60009): You have requested an invalid service instance name. Names are limited to 50 characters.'
```

Would it make sense for App Broker to truncate the service name prefix to ensure that the final name fits within the limit? Or could it fail fast somehow?

royclarkson:
Resolves #204

### DIFF
--- a/spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/targets/TargetServiceTest.java
+++ b/spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/targets/TargetServiceTest.java
@@ -18,10 +18,12 @@ package org.springframework.cloud.appbroker.extensions.targets;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.UUID;
 
 import org.assertj.core.util.Lists;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import reactor.test.StepVerifier;
 
 import org.springframework.cloud.appbroker.deployer.BackingApplication;
 import org.springframework.cloud.appbroker.deployer.TargetSpec;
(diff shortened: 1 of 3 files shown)


spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/extensions/targets/TargetServiceTest.java
```diff
@@ -85,17 +91,64 @@ void shouldAddPropertiesToAllBackingApps() {
...
+				.verifyComplete();
+	}
+
+	private String assembleNameWithGuidSuffix(String appName, String serviceInstanceId) {
+		return appName + CONCAT_DELIM + serviceInstanceId;
+	}
+
+	private int calculateAvailableLength(String serviceInstanceId) {
+		return CF_SERVICE_NAME_MAX_LENGTH - CONCAT_DELIM.length() - serviceInstanceId.length();
+	}
```
LittleBaiBai: This essentially the same logic as in the production code. I personally tend to avoid using code logic to write test, so that we can validate the behavior in two different styles. The exact value don't actually matter, we just want to make sure the name ends up being unique and is less than 50 characters. So maybe we only need to make sure the name includes service instance id, has length of 50, and, optionally, starts with part of the service name. Wdyt?

royclarkson: Yep, I see your point. Thanks! We do need to validate that case where the resulting name is less than 50 chars as well. While I don't like duplicating code, I do consider the goal of the test to be to validate the expected results. Even if the code to do that is basically the same now, the new name still needs to be determined somehow for comparison. But I'll reduce that duplication and update the PR.

LittleBaiBai: Right, it all depends on how we define the goal of this test. 
- If the behavior we are trying to describe here is exactly how we are truncating and concatenate names, then it will be hard to avoid duplicating the logic. 
- If the behavior is just keep it unique and less than 50 chars, then whether it's a `-` in between or how we truncate are not a behavioral requirements that needs to be tested. To achieve this goal, where the resulting name is less than 50 chars, we could assert that it contains the service instance id and the full service name.

royclarkson: ok, redesigned the tests. thanks.

PR #332, opened 2020-01-31
Document example implementation of ServiceInstanceStateRepository

### LINKED ISSUE #257: Update docs to specify that the in memory state repositories should only be used for testing
The in memory state repositories should absolutely not be used for production as they will lose their state if the application restarts. An implementing application should utilize a persistence storage for these state repositories. Related to #254.

### LINKED ISSUE #314: Document that the default behaviour of ServiceInstanceStateRepository
At the moment, we default to an `InMemoryServiceInstanceStateRepository`.
However, we should recommend and provide an example of a better one using a persistence model.

royclarkson:
- Add example implementation in reference docs
- Include warning in javadoc and reference about using in mem implementation
in production

Resolves #257. Resolves #314


royclarkson:
@Albertoimpl I've added the binding reference examples and docs too, for your review.

Issue #337, opened 2020-02-26
Support service instances provisionned by K8S svcat (idempotent dups calls)


gberche-orange:
* As a scab operator
* in order to expose scab-based service brokers to K8S service catalog client
* I need SCAB to implement the concurrency control in the OSB spec 

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#provisioning

> PUT /v2/service_instances/:instance_id
> Response
>
>200 OK | SHOULD be returned if the Service Instance already exists, is fully  provisioned, and the requested parameters are identical to the existing  Service Instance. The expected response body is below.
> -- | --
> 201 Created | MUST be returned if the Service Instance was provisioned as a result of this request. The expected response body is below.
> 202 Accepted | MUST be returned if the Service Instance provisioning is in progress. The operation string MUST match that returned for the original request. This triggers the Platform to poll the Last Operation for Service Instances endpoint for operation status. Note that a re-sent PUT request MUST return a 202 Accepted, not a 200 OK, if the Service Instance is not yet fully provisioned.
> 400 Bad Request | MUST be returned if the request is malformed or missing mandatory  data. MAY be returned if the request contains invalid data, in which  case the error response MAY include a helpful error message in the description field (see Service Broker Errors).
> 409 Conflict | MUST be returned if a Service Instance with the same id already exists or is being provisioned but with different attributes.

See https://github.com/kubernetes-sigs/service-catalog/issues/1639

Currently, duplicate provisionning calls leads to duplicate backend services, or depending on the configuration on errors from Cf reporting already existing app/service instance names

See https://github.com/orange-cloudfoundry/osb-cmdb-spike/issues/17 and https://github.com/orange-cloudfoundry/osb-cmdb-spike/issues/17#issuecomment-589705368 for considered fix using the service instance repository

royclarkson:
@gberche-orange thanks. good catch! I have an unrelated request. Mind if we don't use the acronym SCAB? It's our fault for naming the project that way, but that word has negative connotations in English. :-D Maybe let's just say "app broker"?

gberche-orange:
@royclarkson 
> Mind if we don't use the acronym SCAB?

Sure, I fixed that in the issue body, sorry I did not realize this semantic overriding

royclarkson:
Thanks! much appreciated. No need to apologize. It just reads better for others coming to the project.

PR #343, opened 2020-03-09
K8s service instances dupls (WIP)

gberche-orange:
This is a WIP that cherry picks https://github.com/orange-cloudfoundry/osb-cmdb-spike/pull/27 and starts addressing #337, see related design notes at https://github.com/orange-cloudfoundry/osb-cmdb-spike/issues/17#issuecomment-589705368

Compiles, but there are still some test failures. Need to further diagnose whether this is regression introduced in production code during cherry pick or test that need update. I submitted this PR in case this might help as inspiration, but this is not yet finished.

I'm not sure to be able to allocate much more time on this PR in the short term, so feel free to close this PR if this feels more like pollution.

-----------------

Dupl provisioning, update and unprovisionning request support:
- Modified service instance repository to be checked and saved for "in progress" status on the servlet thread, as to avoid race condition with the reactor thread
- Only skip duplicate request when state repository has an inprogress request, and still fail on async CF-ServiceInstanceNameTaken(60002) for now
- Include log to help diagnose ignored duplicate API calls
- WorkflowServiceInstanceServiceTest completed with a test when a provisionning is in progress

### DIFF
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceService.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceService.java
@@ -25,6 +25,7 @@ import reactor.core.scheduler.Schedulers;
 import reactor.util.Logger;
 import reactor.util.Loggers;
 
+import org.springframework.cloud.appbroker.state.ServiceInstanceState;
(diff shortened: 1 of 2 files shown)


gberche-orange:
This osb-cmdb related feature does not seem to be of interest to app-broker use-cases.  Closing it as I won't be able to rebase/maintain it if there is interest.

Issue #346, opened 2020-03-18
Update CONTRIBUTING docs


Albertoimpl:
The steps to contribute that we have over here https://github.com/spring-cloud/spring-cloud-app-broker/blob/master/CONTRIBUTING.adoc are pointing to a different set of style guides than the ones we require now https://github.com/spring-cloud/spring-cloud-app-broker/tree/master/src

gberche-orange:
thanks @Albertoimpl, sorry I had forgotten about related exchange at https://github.com/spring-cloud/spring-cloud-app-broker/pull/291#issuecomment-551890641

Regarding code formatting including imports, I guess the eclipse code formatted plugin is not necessary anymore and the `src/idea/idea.xml`  needs to be imported in `Editor > code style > import -> intellij idea format`
![image](https://user-images.githubusercontent.com/4748380/76977466-a58ebe00-6935-11ea-9ad1-d8ae99fca2b0.png)

Submitted #347 if this can help.

royclarkson:
Yeah, that's correct. You don't need the eclipse code formatter plugin. That needs to be updated.

spikymonkey:
I think we can close this thanks to #347 - thanks!

Issue #331, opened 2020-01-31
Update javadoc


royclarkson:
Enable pmd rules and add missing javadoc

gberche-orange:
Should include updates to contributing docs, see https://github.com/spring-cloud/spring-cloud-app-broker/pull/347#issuecomment-601432458

PR #347, opened 2020-03-18
Document use of intellij codestyle to format import

gberche-orange:
Document use of intellij codestyle to format import as expected by gradle build

Relates to #346 although more changes might be required such as the followings which do not seem to be currently in sync with project code or PR reviews

> Make sure all new .java files to have a simple Javadoc class comment with at least an @author tag identifying you, and preferably at least a paragraph on what the class is for.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -23,13 +23,8 @@ unacceptable behavior to spring-code-of-conduct@pivotal.io.
 None of these is essential for a pull request, but they will all help.  They can also be
 added after the original pull request but before a merge.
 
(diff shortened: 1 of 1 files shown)


royclarkson:
We have another story to add initial java doc everywhere, and then we may enable the checkstyle rule to validate that on new code. `@author` tags shouldn't be a requirement, so we can remove that from the notes.

gberche-orange:
thanks @royclarkson , cross referencing #331 related to javadoc updates

Albertoimpl:
Thanks for the update @gberche-orange!

PR #352, opened 2020-03-25
Routes property overrides conflicting app manifest properties

spikymonkey:
Currently App Broker allows users to specify mutually exclusive
manifest properties - specifically `routes` along with any combination
of the deprecated `host`, `hosts`, `domain` and `domains`. This change
makes `routes` take precedence if specified.


royclarkson:
@spikymonkey I've cloned and rebased this branch against master to run tests, but the two new tests are failing with NPEs. Looks like we need a `clientApplicationsV3.listPackages` to be mocked as well (maybe more).

spikymonkey:
@royclarkson ah right. The new tests are exercising code added in d724e0dcc4e831120994fe2f52eafa76848514a4 that mustn't have its own tests (otherwise it would also be blowing up). I've added enough mocking to get the tests passing again, but it's not particularly realistic - we should go back and improve the coverage I think. But would this be enough to allow the PR to be merged so I can look at integrating against some snapshot builds?

spikymonkey:
@Albertoimpl this has been merged into the `1.0.x` branch now - can we go ahead and merge to `master` too?

Albertoimpl:
Yep, go ahead! @spikymonkey

spikymonkey:
Cheers!

PR #358, opened 2020-04-01
Run sync binding / unbinding operations synchronously

alek-sys:
Unlike provisioning / deprovisioning, binding can be a fully synchronous
operation, i.e. it should fully complete within a request lifecycle.
There is no need to update service instance state or run create / delete
workflows in a parallel scheduler.

The existing flow should still be enabled for async binding.


royclarkson:
@alek-sys Same comment as the back port PR: Shouldn't we still put a completed or failed status for synchronous binding so last operation will return something?

alek-sys:
I've updated this PR similar to `1.0.x` so sync bindings update state repository.

PR #362, opened 2020-04-07
Run sync binding / unbinding operations synchronously

alek-sys:
Unlike provisioning / deprovisioning, binding can be a fully synchronous
operation, i.e. it should fully complete within a request lifecycle.
There is no need to update service instance state or run create / delete
workflows in a parallel scheduler.

The existing flow should still be enabled for async binding.


alek-sys:
That's a good question. OSBAPI spec for `last_operation` endpoint mentions only provision, update, or deprovision operations. My interpretation of the spec is - nope, sync bindings shouldn't change status of a SI. Instead, sync binding should just complete (binding succesfull) or fail.  Although there is no stanard way to return binding errors then. WDYT?

royclarkson:
It appears the expectation is only for `202 ACCEPTED` async requests, so we should be good. https://github.com/openservicebrokerapi/servicebroker/blob/v2.15/spec.md#polling-last-operation-for-service-bindings

PR #369, opened 2020-05-11
Fix an issue where configured backing services are not being deleted

### LINKED ISSUE #344: Regression: AppDeploymentDeleteServiceInstanceWorkflow leaks backend services on deletion
## Expected behavior

* As an app broker user provisioning brokered service without backing app (i.e. only backing service(s)), see #287
* in order to avoid backing service leak 
* I need backing services to be deleted when a brokered service is unprovisionned

## Observed behavior

Following #318, the AppDeploymentDeleteServiceInstanceWorkflow.deleteBackingServices() method is only attempting to delete backing services bound to applications, and hence skipping backing services when there is no associated backing services.

https://github.com/spring-cloud/spring-cloud-app-broker/blob/b0f2cbc5f4738cb0df8b5d06f3a9888c0188dc06/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java#L75-L92

/CC @LittleBaiBai 

In addition, in case the provisioning completed with error such as the following, then this triggers a leak of backing services
   - the backing services failed to be bound to the backing applications
   - some backing services failed to be created
   - the backing application failed to starter

I understand that the following test enforces this behavior, i.e. that a non bound matching backing service instance does not get deleted upon brokered service deletion
https://github.com/spring-cloud/spring-cloud-app-broker/blob/b0f2cbc5f4738cb0df8b5d06f3a9888c0188dc06/spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/DeleteInstanceWithServicesComponentTest.java#L105-L157

royclarkson:
When a service instance is deleted, all configured backing services will
be deleted as well as all services bound to any backing applications

Resolve #344

### DIFF
--- a/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
+++ b/spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
@@ -16,7 +16,8 @@
 
 package org.springframework.cloud.appbroker.workflow.instance;
 
-import java.util.List;
+import java.util.Collections;
+import java.util.Map;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -27,6 +28,7 @@ import org.springframework.cloud.appbroker.deployer.BackingAppDeploymentService;
 import org.springframework.cloud.appbroker.deployer.BackingService;
 import org.springframework.cloud.appbroker.deployer.BackingServicesProvisionService;
 import org.springframework.cloud.appbroker.deployer.BrokeredServices;
+import org.springframework.cloud.appbroker.deployer.DeploymentProperties;
 import org.springframework.cloud.appbroker.extensions.credentials.CredentialProviderService;
 import org.springframework.cloud.appbroker.extensions.targets.TargetService;
 import org.springframework.cloud.appbroker.manager.BackingAppManagementService;
@@ -35,6 +37,7 @@ import org.springframework.cloud.servicebroker.model.instance.DeleteServiceInsta
 import org.springframework.cloud.servicebroker.model.instance.DeleteServiceInstanceResponse;
 import org.springframework.cloud.servicebroker.model.instance.DeleteServiceInstanceResponse.DeleteServiceInstanceResponseBuilder;
 import org.springframework.core.annotation.Order;
+import org.springframework.util.CollectionUtils;
 
 @Order(0)
 public class AppDeploymentDeleteServiceInstanceWorkflow
@@ -75,11 +78,14 @@ public class AppDeploymentDeleteServiceInstanceWorkflow
 	}
 
 	private Flux<String> deleteBackingServices(DeleteServiceInstanceRequest request) {
-		return collectConfiguredBackingServices(request)
-			.mergeWith(collectBoundBackingServices(request))
-			.doOnEach(backingServices -> log.debug("Deleting backing services {} for {}/{}",
-				backingServices, request.getServiceDefinition().getName(), request.getPlan().getName()))
-			.flatMap(backingServicesProvisionService::deleteServiceInstance)
+		return collectBackingServices(request)
+			.collectList()
+			.flatMapMany(backingServices -> {
+				if (!CollectionUtils.isEmpty(backingServices)) {
+					return backingServicesProvisionService.deleteServiceInstance(backingServices);
+				}
+				return Flux.empty();
+			})
 			.doOnComplete(() -> log.debug("Finished deleting backing services for {}/{}",
 				request.getServiceDefinition().getName(), request.getPlan().getName()))
 			.doOnError(exception -> log.error(String.format("Error deleting backing services for %s/%s with error '%s'",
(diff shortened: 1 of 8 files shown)


spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/DeleteInstanceWithServicesComponentTest.java
```diff
@@ -30,26 +30,22 @@
...
-	"spring.cloud.appbroker.services[0].apps[0].name=" + APP_NAME,
-	"spring.cloud.appbroker.services[0].apps[0].services[0].service-instance-name=" + BACKING_SI_NAME,
 	"spring.cloud.appbroker.services[0].services[0].service-instance-name=" + BACKING_SI_NAME,
 	"spring.cloud.appbroker.services[0].services[0].name=" + BACKING_SERVICE_NAME,
 	"spring.cloud.appbroker.services[0].services[0].plan=" + BACKING_PLAN_NAME
 })
 class DeleteInstanceWithServicesComponentTest extends WiremockComponentTest {
 
-	protected static final String APP_NAME = "app-delete-with-services";
+//	protected static final String APP_NAME = "app-delete-with-services";
```
LittleBaiBai, line 48: Commented line to remove.

royclarkson: Thanks!

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
```diff
@@ -73,14 +75,8 @@ public AppDeploymentDeleteServiceInstanceWorkflow(BrokeredServices brokeredServi
...
-		return backingAppManagementService.getDeployedBackingApplications(request.getServiceInstanceId())
-			.flatMapMany(Flux::fromIterable)
-			.flatMap(backingApplication ->
-				Flux.fromIterable(backingApplication.getServices())
-					.map(servicesSpec -> BackingService.builder()
-						.serviceInstanceName(servicesSpec.getServiceInstanceName())
-						.build())
-					.collectList())
+		return collectConfiguredBackingServices(request)
+			.mergeWith(collectBoundBackingServices(request))
```
gberche-orange, line 79: Would there be duplicate BackingServices among those returned by `collectConfiguredBackingServices` and `collectBoundBackingServices` ?

Does this lead to polluting traces when trying to delete twice a service instance ?

royclarkson: Thanks for the feedback. Yes, it would result in multiple delete requests if a backing service instance is bound to an app. I'll consider how to eliminate those duplicates.

royclarkson: I've added a new commit that addresses the issue of duplicate service instances

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
```diff
@@ -75,34 +77,52 @@ public AppDeploymentDeleteServiceInstanceWorkflow(BrokeredServices brokeredServi
...
+			.flatMap(backingApplication -> Mono.justOrEmpty(backingApplication.getServices())
+				.flatMapMany(Flux::fromIterable)
+				.flatMap(servicesSpec -> Mono.justOrEmpty(servicesSpec.getServiceInstanceName()))
+				.map(serviceInstanceName -> {
+					BackingService backingService = BackingService.builder()
+						.serviceInstanceName(serviceInstanceName)
+						.build();
+					if (!CollectionUtils.isEmpty(backingApplication.getProperties())) {
+						backingService.setProperties(Collections.singletonMap(DeploymentProperties.TARGET_PROPERTY_KEY,
+							backingApplication.getProperties().get(DeploymentProperties.TARGET_PROPERTY_KEY)));
```
LittleBaiBai, line 122: This line is pretty long and hard to follow. Would it be worth breaking the get target part out? Also, if the target property is null on backing application, we can probably skip setting it on backing service.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
```diff
@@ -75,34 +77,52 @@ public AppDeploymentDeleteServiceInstanceWorkflow(BrokeredServices brokeredServi
...
-					.serviceInstanceName(servicesSpec.getServiceInstanceName())
-					.build())
-				.collectList());
+			.flatMap(backingApplication -> Mono.justOrEmpty(backingApplication.getServices())
+				.flatMapMany(Flux::fromIterable)
+				.flatMap(servicesSpec -> Mono.justOrEmpty(servicesSpec.getServiceInstanceName()))
+				.map(serviceInstanceName -> {
+					BackingService backingService = BackingService.builder()
+						.serviceInstanceName(serviceInstanceName)
+						.build();
```
LittleBaiBai, line 119: Nit-picking: I would personally prefer moving the `.build()` to the return line and avoid using setter on backing service. It would give me a signal that this whole section is about building backing service. But the way it is right now is totally fine too.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
```diff
@@ -75,34 +77,52 @@ public AppDeploymentDeleteServiceInstanceWorkflow(BrokeredServices brokeredServi
...
-			.doOnEach(backingServices -> log.debug("Deleting backing services {} for {}/{}",
-				backingServices, request.getServiceDefinition().getName(), request.getPlan().getName()))
-			.flatMap(backingServicesProvisionService::deleteServiceInstance)
+		return collectBackingServices(request)
+			.collectList()
+			.flatMapMany(backingServices -> {
+				if (!CollectionUtils.isEmpty(backingServices)) {
+					return backingServicesProvisionService.deleteServiceInstance(backingServices);
+				}
+				return Flux.empty();
```
LittleBaiBai, line 86: `Flux.empty().collectList()` will result in an empty list instead of null, so I feel this null check here is not necessary.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
royclarkson: I'll rework it. Thanks!

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
royclarkson: I got a NPE in the mapper without that check. I agree it's ugly and feels unnecessary. :-/

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/workflow/instance/AppDeploymentDeleteServiceInstanceWorkflow.java
royclarkson: makes sense. I'll move that logic to another function to make it read better.

royclarkson: reworked instead of moving to another function

PR #373, opened 2020-05-18
Add new logging starter and refactor logging autoconfiguration

royclarkson:
See 'spring-cloud-starter-app-broker-logging'


spring-cloud-starter-app-broker-logging/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+dependencyManagement {
+	imports {
+		mavenBom "org.springframework.boot:spring-boot-dependencies:${springBootVersion}"
+	}
+}
+
+dependencies {
+	compile project(":spring-cloud-starter-app-broker")
+	compile project(":spring-cloud-app-broker-logging")
```
LittleBaiBai, line 27: Could we use `api` instead of `compile`?

royclarkson: I'm going to refactor the entire gradle build to use the new maven publishing plugin soon and I'll change everything to api/integration at that point.

royclarkson:
yes, that's the correct workflow for how to enable logging support

PR #374, opened 2020-05-20
Remove CAPI call to get service name and plan

alek-sys:
Calling /v2/service_instance/<id> requires the broker client to be
either space developer of a user's space or have `cloud_controller.admin`
authority. To remove this requirement, BackingAppManagementService was
updated to accept service and plan name as arguments, so now getting
these values is a responsibility of a caller. In case of update / delete
both values can be extracted from the request, for management operations
they now have to be provided.

Also, this commit introduces ATs refactoring to make sure App Broker can
work without `cc.admin` permissions and without being a org manager.
This is done by separating CloudFoundryService to User and Broker
implementations. UserCloudFoundryService can only create / update /
delete service instances in own space and org, which broker UAA client
has no access to. BrokerCloudFoundryService can work with the backing
apps, services, spaces etc.

### DIFF
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
@@ -66,6 +66,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration;
 import org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryService;
+import org.springframework.cloud.appbroker.acceptance.fixtures.cf.UserCloudFoundryService;
 import org.springframework.cloud.appbroker.acceptance.fixtures.uaa.UaaService;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.client.reactive.ReactorClientHttpConnector;
@@ -76,10 +77,14 @@ import org.springframework.web.reactive.function.client.WebClient;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration.APP_BROKER_CLIENT_AUTHORITIES;
 import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration.APP_BROKER_CLIENT_SECRET;
+import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration.USER_CLIENT_AUTHORITIES;
+import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration.USER_CLIENT_ID;
+import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration.USER_CLIENT_SECRET;
 
 @SpringBootTest(classes = {
 	CloudFoundryClientConfiguration.class,
 	CloudFoundryService.class,
+	UserCloudFoundryService.class,
 	UaaService.class,
 	HealthListener.class,
 	RestTemplate.class
@@ -105,6 +110,9 @@ abstract class CloudFoundryAcceptanceTest {
 
 	@Autowired
 	protected CloudFoundryService cloudFoundryService;
+
+	@Autowired
+	protected UserCloudFoundryService userCloudFoundryService;
 
 	@Autowired
 	private UaaService uaaService;
@@ -136,6 +144,7 @@ abstract class CloudFoundryAcceptanceTest {
 	void setUp(TestInfo testInfo, BrokerProperties brokerProperties) {
 		List<String> appBrokerProperties = getAppBrokerProperties(brokerProperties);
 		blockingSubscribe(initializeBroker(appBrokerProperties));
+		blockingSubscribe(initializeUser());
 	}
 
 	void setUpForBrokerUpdate(BrokerProperties brokerProperties) {
@@ -197,16 +206,29 @@ abstract class CloudFoundryAcceptanceTest {
 
 	@AfterEach
 	public void tearDown(TestInfo testInfo) {
-		blockingSubscribe(cloudFoundryService.getOrCreateDefaultOrganization()
+		blockingSubscribe(cloudFoundryService.getOrCreateDefaultOrg()
 			.map(OrganizationSummary::getId)
 			.flatMap(orgId -> cloudFoundryService.getOrCreateDefaultSpace()
 				.map(SpaceSummary::getId)
 				.flatMap(spaceId -> cleanup(orgId, spaceId))));
 	}
 
+	private Mono<Void> initializeUser() {
+		return cloudFoundryService.getOrCreateOrganization(userCloudFoundryService.getOrgName())
+			.map(OrganizationSummary::getId)
+			.flatMap(orgId -> cloudFoundryService
+				.getOrCreateSpace(userCloudFoundryService.getOrgName(), userCloudFoundryService.getSpaceName())
+				.map(SpaceSummary::getId)
+				.flatMap(spaceId -> uaaService.createClient(
+						USER_CLIENT_ID,
+						USER_CLIENT_SECRET,
+						USER_CLIENT_AUTHORITIES)
+					.then(cloudFoundryService
+						.associateClientWithOrgAndSpace(USER_CLIENT_ID, orgId, spaceId))));
+	}
+
 	private Mono<Void> initializeBroker(List<String> appBrokerProperties) {
-		return cloudFoundryService
-			.getOrCreateDefaultOrganization()
+		return cloudFoundryService.getOrCreateDefaultOrg()
 			.map(OrganizationSummary::getId)
 			.flatMap(orgId -> cloudFoundryService
 				.getOrCreateDefaultSpace()
@@ -216,7 +238,8 @@ abstract class CloudFoundryAcceptanceTest {
 						brokerClientId(),
 						APP_BROKER_CLIENT_SECRET,
 						APP_BROKER_CLIENT_AUTHORITIES))
-					.then(cloudFoundryService.associateAppBrokerClientWithOrgAndSpace(brokerClientId(), orgId, spaceId))
+					.then(cloudFoundryService
+						.associateAppBrokerClientWithOrgAndSpace(brokerClientId(), orgId, spaceId))
 					.then(cloudFoundryService
 						.pushBrokerApp(testBrokerAppName(), getTestBrokerAppPath(), brokerClientId(),
 							appBrokerProperties))
@@ -250,7 +273,7 @@ abstract class CloudFoundryAcceptanceTest {
 		String planName,
 		String serviceInstanceName,
 		Map<String, Object> parameters) {
-		cloudFoundryService.createServiceInstance(planName, serviceName, serviceInstanceName, parameters)
+		userCloudFoundryService.createServiceInstance(planName, serviceName, serviceInstanceName, parameters)
 			.then(getServiceInstanceMono(serviceInstanceName))
 			.flatMap(serviceInstance -> {
 				assertThat(serviceInstance.getStatus())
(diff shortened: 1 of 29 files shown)


spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/fixtures/cf/BrokerCloudFoundryService.java
```diff
@@ -330,42 +304,48 @@ public CloudFoundryService(CloudFoundryClient cloudFoundryClient,
...
 				.build())
-				.then(getDefaultSpace(spaceOperations)));
+				.then(getSpace(spaceOperations, spaceName)));
+	}
+
+	public String getDefaultSpaceName() {
+		return cloudFoundryProperties.getDefaultSpace();
 	}
 
-	public Mono<OrganizationSummary> getOrCreateDefaultOrganization() {
```
royclarkson, line 343: Should we leave this one in place too?

spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/manager/BackingAppManagementServiceTest.java
```diff
@@ -135,75 +125,35 @@ void stopApplicationsWithEmptyApplications() {
...
 			.verifyComplete();
 
-		verify(appDeployer).getServiceInstance(any(GetServiceInstanceRequest.class));
 		verify(targetService).addToBackingApplications(eq(emptyBackingApps), any(), eq("foo-service-id"));
 		verifyNoInteractions(managementClient);
 		verifyNoMoreInteractions(appDeployer, targetService, managementClient);
 	}
 
-	@Test
-	void stopApplicationsServiceNotFound() {
```
royclarkson, line 158: Is there not any value in these negative test cases?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -197,44 +202,47 @@ public MappingProvider mappingProvider() {
...
-			.getOrCreateDefaultOrganization()
+		return brokerCloudFoundryService
+			.getOrCreateOrganization(userCloudFoundryService.getOrgName())
+			.then(brokerCloudFoundryService.getOrCreateSpace(userCloudFoundryService.getSpaceName()))
+			.then(brokerCloudFoundryService.getOrCreateOrganization(brokerCloudFoundryService.getDefaultOrgName()))
 			.map(OrganizationSummary::getId)
-			.flatMap(orgId -> cloudFoundryService
-				.getOrCreateDefaultSpace()
+			.flatMap(orgId -> brokerCloudFoundryService
+				.getOrCreateSpace(brokerCloudFoundryService.getDefaultSpaceName())
```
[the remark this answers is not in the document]
alek-sys: My intention was to move all admin operations into Admin service and parametrize them. But I don't mind keeping `getOrCreateDefaultSpace/Org` methods as shortcuts as you suggested below, it is certainly making code more readable.

spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/manager/BackingAppManagementServiceTest.java
alek-sys: Maybe in a different form. Without making a call to CAPI, broker doesn't know if an SI exists or not. So if a management operation invoked for a non-existing service instance, the operation will fail with either non-existing space exception (wtih space-per-si target) or non-existing application instance (with space guid app suffix target). But in any case App Broker will just relay CAPI exception to a caller, so not sure how much value is in testing this. 

Another option I can think of is handling non-existing space / app exception and wrapping that to `ServiceInstanceOperationException` or something. But even then, App Broker can only _guess_ that if a space is not found then SI doesn't exist, there could be other reasons as well. 

So I'm happy to bring this negative test cases back, but I'm not sure what exactly they suppose to be testing. WDYT?

spikymonkey:
Hey @alek-sys, I'm struggling a bit with this PR. I can't help but think that it's quite a bit of extra complexity - especially around the mixed admin/user services. I understand that we can't make the CAPI call to fetch a user's service instance by id (although still haven't dug far enough back to understand what has changed around that between 1.0.x and 1.1.x). But that just makes me feel that there might be a gap in our own state tracking requirements in app broker. Also I'm wondering if, with some work around `Target`, we might be able to determine backing apps from service instance id alone without making the problematic CAPI call. Wdyt?

spikymonkey:
Thanks @alek-sys and @royclarkson for taking me through this. Just to capture the context here:

* We've agreed to maintain the existing `BackingAppManagementService` public API which allows management operations using just a service instance id. We'll add Javadoc to make it clear that this will only worked if invoked using a broker client which has the `cloud_controller.admin` authority.
* We'll also overload those management operations with methods which take the service and plan names which are required if they're being invoked by clients which _don't_ have the `cloud_controller.admin` authority. This gives implementors the choice to use less privileged clients if they want to.
* We'll update the ATs to use a distinct user for the `UserCloudFoundryService`, rather than sharing the existing broker client. This brings the tests into line with real use cases and should help to prevent any future cases where we find the tests are able to perform operations that a real user or broker could not.

royclarkson:
@alek-sys where are we with this PR? I see you pushed a few changes. Any more work required, or is it ready to be reviewed again?

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/manager/BackingAppManagementService.java
```diff
@@ -82,8 +114,24 @@ public BackingAppManagementService(ManagementClient managementClient, AppDeploye
 			.then();
 	}
 
+	/**
+	 * Helper method that fetches service name and plan name from Cloud Foundry Service Instances API (CF API) and
+	 * invokes {@code restart(serviceInstanceId, serviceName, planName)}.
+	 *
+	 * Because this method will try to fetch user-created service instance, UAA client used by the broker has to be a
+	 * space developer and an org user of the target space and org, or has to have {@code cloud_controller.admin}
```
spikymonkey, line 122: could 'target space and org' be confused with the backing app target space and org? Also tempted to drop the org user bit since it's implied by space developer. Wonder if we should reword slightly to something like 'the space containing the service instance'

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/manager/BackingAppManagementService.java
```diff
@@ -82,8 +114,24 @@ public BackingAppManagementService(ManagementClient managementClient, AppDeploye
 			.then();
 	}
 
+	/**
+	 * Helper method that fetches service name and plan name from Cloud Foundry Service Instances API (CF API) and
+	 * invokes {@code restart(serviceInstanceId, serviceName, planName)}.
```
spikymonkey, line 119: By referencing e.g. `restart(serviceInstanceId, serviceName, planName)` in this Javadoc it makes me think we should Javadoc those methods too. Just something simple like 'Restarts the backing applications for the service instance with the given id'.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -104,7 +106,10 @@
 	protected static final String BACKING_APP_PATH = "classpath:backing-app.jar";
 
 	@Autowired
-	protected CloudFoundryService cloudFoundryService;
+	protected AdminCloudFoundryService adminCloudFoundryService;
```
spikymonkey, line 109: Sorry, I'd like to rename this back to `BrokerCloudFoundryService` or something similar. It is the client that represents the broker, and doesn't have any admin authorities.

Also, can it be made `private`? I haven't checked to see all the use cases yet - just a thought. I'm thinking that all tests should be written from unprivileged user perspective rather than using the broker client, which wasn't an option before this PR.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -136,6 +141,7 @@ private String brokerClientId() {
 	void setUp(TestInfo testInfo, BrokerProperties brokerProperties) {
 		List<String> appBrokerProperties = getAppBrokerProperties(brokerProperties);
 		blockingSubscribe(initializeBroker(appBrokerProperties));
+		blockingSubscribe(initializeClient());
```
spikymonkey, line 144: We should be explicit which client this is, since we have two of them and it's not immediately obvious what's going on here. Ideally I think we'd use a user here instead of a client.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/fixtures/cf/UserCloudFoundryService.java
```diff
@@ -0,0 +1,141 @@
...
+import reactor.core.publisher.Mono;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public final class UserCloudFoundryService {
+
+	private static final Logger LOG = LoggerFactory.getLogger(UserCloudFoundryService.class);
+
+	private static final String USER_CLIENT_ID = "app-broker-test-client";
```
spikymonkey, line 41: It's a bit surprising that these are here, along with the authorities a bit further down, but the equivalent configuration for `AdminCloudFoundryService` is in `CloudFoundryClientConfiguration`. Can we make this more consistent?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
alek-sys: Sure, renamed it back to `BrokerCloudFoundryService`. Now thinking it could be not the best name since this service uses client id and secret provided via properties, which most likely is admin credentials and it does not represent the broker. So it is really `TestAdminCloudFoundryService`. But I think it is not an issue so happy to leave it `BrokerCloudFoundryService`. 

We still use this class in management tests, so can't yet make it `private`. I don't really see it a problem because these two classes have different methods, so it's not easy to mix them up.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/fixtures/cf/UserCloudFoundryService.java
alek-sys: Yep, sorry I missed that. Not sure I like having all these constants in `CloudFoundryClientConfiguration`, but I agree it's better to have it consitent.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
spikymonkey: > this service uses client id and secret provided via properties, which most likely is admin credentials

Sorry @alek-sys I see what you mean now. I'd misunderstood which clients were in play here.  So there are three... the 'admin' client (used by tests) which is configured via properties, and then the 'broker' (used by the broker under test) and 'user' (used by the tests) clients which are configured dynamically in the test setup. Have I finally understood this properly? 😅 

So calling the 'admin' client the 'broker' client is misleading. Now I think you're right about `TestAdminCloudFoundryService`. Sorry, sorry...!

I think the most confusing thing is how we seem to flip between usage of the 'admin' and 'user' clients, e.g.

```
	protected Mono<String> manageApps(String serviceInstanceName, String serviceName,
		String planName, String operation) {
		return userCloudFoundryService
			.getServiceInstance(serviceInstanceName)
			.map(ServiceInstance::getId)
			.flatMap(serviceInstanceId ->
				adminCloudFoundryService
					.getApplicationRoute(testBrokerAppName())
					.flatMap(appRoute ->
```

but at least I think I understand why now. @royclarkson made a comment about merging the `admin` and `user` clients together - what do you think about this? Now that I'm understanding that these clients both represent the tests, and that the 'broker' client (the one actually representing the broker) does not have `cloud_controller.admin`, I quite like that idea. Would be less confusing? If the tests use their admin authority to access SI's in the user's org and space, in order to drive a broker which does _not_ have admin authority, are we okay with that? Or am I missing a problem?

PR #378, opened 2020-06-02
Stop log streaming if listener is dropped

alek-sys:
CLI plugin doesn't always handle disconnect gracefully, so sometimes it is possible that log stream is being published but there is no listener. In this case log streaming should be stopped.

### DIFF
--- a/spring-cloud-app-broker-logging/src/test/java/org/springframework/cloud/appbroker/logging/streaming/ServiceInstanceLogStreamingTest.java
+++ b/spring-cloud-app-broker-logging/src/test/java/org/springframework/cloud/appbroker/logging/streaming/ServiceInstanceLogStreamingTest.java
@@ -40,10 +40,13 @@ import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.cloud.appbroker.logging.LoggingUtils;
+import org.springframework.cloud.appbroker.logging.streaming.events.ServiceInstanceLogEvent;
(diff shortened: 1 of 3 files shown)


spring-cloud-app-broker-logging/src/test/java/org/springframework/cloud/appbroker/logging/streaming/ServiceInstanceLogStreamingTest.java
```diff
@@ -134,6 +140,19 @@ void shouldPublishEventOnDisconnect() {
...
+	@Test
+	void shouldStopStreamingIfNoClient() {
+		// CLI plugin doesn't always handle disconnect gracefully, so sometimes it is possible that log stream is
+		// being published but there is no listener. In this case log streaming should be stopped.
+
+		Disposable subscription = connectToLogsStreamEndpoint();
+		subscription.dispose();
+
+		applicationEventPublisher.publishEvent(new ServiceInstanceLogEvent(this, serviceInstanceId, expectedEnvelope));
+		await().atMost(Duration.ofSeconds(1))
```
spikymonkey, line 152: I just saw a failure in CI because of a timeout in an 'at most 1 seconds' test. Do we really want to be this aggressive? What runs quickly on a dev box may not run so quickly on a worker node...

alek-sys: I'm happy to increase it, but if RSocket message is not published locally within 1 second I'd be concerned. Also increasing timeout will increase test run time if there is a real issue with code so the message doesn't get published. Are you sure this CI failure was a flake due to agressive timeout?

spikymonkey: Just something I'm a bit wary of given recent build failures. We're implicitly testing performance of environments we don't control here as well as correctness. I would totally hope we'd get our messages through within a second too, but equally I don't really mind if they take 5 or 10 because the env is having a hiccup or something.

PR #386, opened 2020-06-08
Updates to the staging process

spikymonkey:
* Pipeline now fetches tags so the task doesn't have to
* Staging jobs now take git versions that have passed ATs. This avoids
the need to run all the tests again, taking staging down from ~20 mins
to ~2
* Version bumps are done on a branch to avoid pushing to protected
`master branch`
* Debug output and script tidying


royclarkson:
rebased with master and closed in f2ec97d97773ced4068d23a2f15909a01a0e5067

PR #395, opened 2020-07-01
Use standard gradle maven plugins if possible

### LINKED ISSUE #360: Use standard Gradle maven plugins if possible
I just tried to `./gradlew publishToMavenLocal` and realised that we're not using the standard Gradle maven publishing plugin. I'm wondering if we can? Could be an interesting spike to see whether there are still gaps between the standard tooling and our own.

### LINKED ISSUE #382: Upgrade to Spring Boot 2.3

### LINKED ISSUE #391: Prevent the root Gradle project from producing a POM
Maven Central won't accept this POM, so it has to be manually deleted before syncing. We should just stop generating it.
Related to #360

spikymonkey:
closes #360, closes #391, closes #382


spikymonkey:
Still need to look at OSB update and running the ATs on this one, but hopefully getting to the end of this now. And also understand the build a lot better than I did 😄

spikymonkey:
@royclarkson another thing for review - I used the native `distribution` plugin to package up the dist and docs, as I thought it was a bit more 'the Gradle way' than writing custom tasks. But a side-effect of that is that the zip files that are produced have a slightly different naming format. It's possible to tweak the naming to some degree, but not possible to get exactly what you had before, because it insists on tacking `version` onto the end of the filename. I wasn't sure if that matters or not. If it does, totally happy to go back to what we had before.

spring-cloud-app-broker-docs/build.gradle
```diff
@@ -31,22 +25,17 @@ configurations {
...
-	implementation("org.springframework.boot:spring-boot-starter-security")
-	implementation("org.springframework.boot:spring-boot-starter-tomcat")
-	implementation("io.projectreactor:reactor-core")
-	implementation("org.springframework.boot:spring-boot-starter-data-r2dbc:2.3.0.RELEASE")
-	implementation("io.r2dbc:r2dbc-h2:0.8.0.RELEASE")
+	implementation "org.springframework.boot:spring-boot-starter"
+	implementation "org.springframework.boot:spring-boot-starter-security"
+	implementation "org.springframework.boot:spring-boot-starter-tomcat"
+	implementation "io.projectreactor:reactor-core"
+	implementation "org.springframework.data:spring-data-r2dbc:1.1.1.RELEASE"
```
royclarkson, line 35: Can we not use the boot r2dbc starter from boot 2.3?

spring-cloud-starter-app-broker-logging/build.gradle
```diff
@@ -16,21 +16,7 @@
...
-
 dependencies {
-	compile project(":spring-cloud-starter-app-broker")
-	compile project(":spring-cloud-app-broker-logging")
-}
-
-install {
-	repositories.mavenInstaller {
-		pom.whenConfigured { generatedPom ->
-			generatedPom.packaging = "pom"
```
royclarkson, line 33: I had to do the following in SCOSB to get it to set the pom packaging on the starters:
```
generatePomFileForMavenJavaPublication {
	pom.packaging = "pom"
}
```

spring-cloud-starter-app-broker/build.gradle
```diff
@@ -16,22 +16,8 @@
...
 dependencies {
-	compile project(":spring-cloud-app-broker-core")
-	compile project(":spring-cloud-app-broker-autoconfigure")
-	compile("org.springframework.cloud:spring-cloud-starter-open-service-broker:${openServiceBrokerVersion}")
-}
-
-install {
-	repositories.mavenInstaller {
-		pom.whenConfigured { generatedPom ->
-			generatedPom.packaging = "pom"
```
royclarkson, line 34: same as previous comment

```
generatePomFileForMavenJavaPublication {
	pom.packaging = "pom"
}
```

spring-cloud-starter-app-broker-cloudfoundry/build.gradle
```diff
@@ -16,24 +16,11 @@
...
+	api project(":spring-cloud-app-broker-deployer-cloudfoundry")
+	api project(":spring-cloud-app-broker-security-credhub")
+	api "org.cloudfoundry:cloudfoundry-client-reactor:${cfJavaClientVersion}"
+	api "org.cloudfoundry:cloudfoundry-operations:${cfJavaClientVersion}"
 }
 
-install {
-	repositories.mavenInstaller {
-		pom.whenConfigured { generatedPom ->
-			generatedPom.packaging = "pom"
```
royclarkson, line 36: Consider the following to set pom packaging:
```
generatePomFileForMavenJavaPublication {
	pom.packaging = "pom"
}
```

build.gradle
```diff
@@ -225,130 +172,204 @@ configure(subprojects - [project(":spring-cloud-starter-app-broker"),
...
+		publications {
+			mavenJava(MavenPublication) {
+				suppressAllPomMetadataWarnings()
+				from components.java
+				pom {
+					afterEvaluate {
+						name = project.description
+						description = project.description
+					}
+					packaging = starterProjects.contains(project) ? "pom" : null
```
royclarkson, line 246: oh I see. you can ignore my other comments about pom packaging in the starters if this works as I'm reading it.

royclarkson:
@spikymonkey sorry, I'm not clear on the details of how those artifacts are published from artifactory. The only thing we configure is this [bit](https://github.com/spring-cloud/spring-cloud-app-broker/blob/master/ci/pipeline.yml#L46-L51) about the docs zip. Does it still match the new file format?

ci/scripts/stage.sh
```diff
@@ -37,7 +37,7 @@ git commit -m "Release v$stageVersion"
 git tag -a "v$stageVersion" -m "Release v$stageVersion"
 echo
 
-./gradlew --no-daemon clean install -Dmaven.repo.local="${repository}"
+./gradlew clean publish -PpublicationRepository="${repository}"
```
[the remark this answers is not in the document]
spikymonkey: Gradle's recommendation since 3.0 is to [use the daemon in CI](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:disabling_the_daemon) unless for some reason it's causing instability.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5.1-all.zip
```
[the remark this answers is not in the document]
spikymonkey: Hmm. I changed this because it gives IntelliJ a lot of help when working on the build scripts (it'll nag you to switch by default). There'll be a small download time penalty in CI, yes. Probably only a few seconds... worth it?

settings.gradle
```diff
@@ -1,3 +1,14 @@
+pluginManagement {
+	plugins {
+		id "io.spring.dependency-management" version "1.0.9.RELEASE"
+		id "org.springframework.boot" version "2.2.7.RELEASE"
+		id "com.jfrog.artifactory" version "4.15.2"
```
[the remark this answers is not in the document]
spikymonkey: Oh good catch, I missed this. I'm super confused by that artifactory dependency... I'd thought I'd copied this from `master` but apparently not. Seems like a copy/paste mistake. The Boot version should also be parametereised and dependency-management can go.

spring-cloud-app-broker-docs/build.gradle
spikymonkey: We can. I hit something that caused me to change this, but now things have settled again, the starter works just fine.

spikymonkey:
Thanks @royclarkson, another good catch. I hadn't spotted that docs glob in the pipeline. I don't love the way that the artifactory publication is split between the gradle build and pipeline - doing half the job in one and the rest in the other makes it a bit hard to follow. The other important bit in the pipeline config is `folder: distribution-repository` - we do a local repo install to that folder and then pull the artifacts from there, along with the docs zip that you spotted. It might be worth investigating the artifactory plugin to see if pulling all of this into the build makes things simpler, but I wonder if that's a job for another time.

ci/scripts/stage.sh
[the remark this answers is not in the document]
royclarkson: thanks for the explanation!

spring-cloud-app-broker-docs/build.gradle
royclarkson: I just realized we haven't bumped the project to boot 2.3 yet, so I'm not sure it does make sense to mix that version in with the boot 2.2 dependencies. Probably best to leave it as is, and then update these dependencies when the project is upgraded.

spikymonkey: I have updated to boot 2.3 :)

Issue #403, opened 2020-11-26
Fix pipeline failures due to DockerHub rate limiting


spikymonkey:
The App Broker Concourse pipelines are failing due to our unauthenticated image pulls hitting the new DockerHub rate limits. We need to work around this, either by hosting the CI base images in a private registry or using an authenticated account to do the pull.

PR #404, opened 2020-11-27
Replace docker-image resource with registry-image

spikymonkey:
This can make use of the global Concourse registry mirror, which is
authenticated to avoid DockerHub rate limits.

#403


dyroberts (approved):
Lasgun
Gom jabbar
Thopter
Melange

PR #405, opened 2020-11-30
Pass through pipeline vars to tasks that require them

spikymonkey:


royclarkson (approved):
LGTM, and we know it works already.

spikymonkey:
Thanks 👍  I updated the job grouping as we discussed yesterday too - less easy to accidentally click the wrong one now 😄

PR #407, opened 2021-01-11
Upgrade to Spring Boot 2.4.2

royclarkson:
- Upgrade to Spring Cloud Open Service Broker 3.3.0-RC1
- Resolve deprecation warnings


spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
```diff
@@ -139,7 +139,7 @@ else if (isRouteBindingRequest(request)) {
 
 	private boolean isAppBindingRequest(CreateServiceInstanceBindingRequest request) {
 		return request.getBindResource() != null
-			&& StringUtils.isEmpty(request.getBindResource().getRoute());
+			&& !StringUtils.hasLength(request.getBindResource().getRoute());
```
spikymonkey, line 142: What's the reason for this change? I think `isEmpty` is easier to parse than negated `hasLength`.

spikymonkey (approved):
Thanks @royclarkson!

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
spikymonkey: Oh, now I see I was looking at the old version of `StringUtils` and `isEmpty` is deprecated as of 5.3.

royclarkson: Yeah, I prefer the `isEmpty` API as well. It's very clear to me. I guess there were reasons for deprecating it.

spring-cloud-app-broker-autoconfigure/src/main/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceLogStreamAutoConfiguration.java
```diff
@@ -59,7 +59,7 @@ public WebSocketHandlerAdapter handlerAdapter() {
 	@Bean
 	public HandlerMapping logsHandlerMapping(StreamingLogWebSocketHandler webSocketHandler) {
 		Map<String, WebSocketHandler> map = new HashMap<>();
-		map.put("/logs/**/stream", webSocketHandler);
+		map.put("/logs/**", webSocketHandler);
```
royclarkson, line 62: @spikymonkey this is the change that I don't like. Boot changed the default parser and it doesn't support wildcards in the middle of the template like that anymore. You can set a config option to reenable the ant style parser, but I'd rather not do that. All the tests pass, which I guess means it's still working as desired. 😅

spikymonkey: Looking again at this, I found this logging pattern in.  `StreamingLogWebSocketHandler`:

```
private static final UriTemplate LOGGING_URI_TEMPLATE = new UriTemplate("/logs/{serviceInstanceId}/stream");
```

Is that the only pattern we need to match? Assuming we don't need to match things like `/logs/multiple/path/segments/stream` then we don't need the Ant style wildcards.

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
spikymonkey: You could use `ObjectUtils.isEmpty` instead.

spring-cloud-app-broker-autoconfigure/src/main/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceLogStreamAutoConfiguration.java
royclarkson: That's the only one for which I'm aware. 🤷‍♂️

spring-cloud-app-broker-core/src/main/java/org/springframework/cloud/appbroker/service/WorkflowServiceInstanceBindingService.java
royclarkson: Good point. I'm good either way, but the isEmpty is still more clear to me.

Albertoimpl (approved):

PR #408, opened 2021-01-20
Pre-update of backing applications

spikymonkey:
The update service flow is now:
1. pre-update backing applications
2. update backing services
3. update backing applications

The Cloud Foundry implementation has a pre-update step which applies
backing application environment updates. This allows the following
backing services update step to read the updated app environment, which
is required for correct operation of some services where a rebind is
performed on update. The final update backing applications step performs
a rolling redeployment as before to ensure both environment changes and
any credential updates resulting from a rebind are propagated to the
backing apps.


LittleBaiBai (approved):
LGTM, and the AT passing of my open SCG4TAS PR is a good sign 😄 .

spikymonkey:
Thanks for the review @LittleBaiBai 👍

PR #409, opened 2021-02-18
Upgrading CFJC to 5.1

Albertoimpl:
[Connected to #406]


README.adoc
```diff
@@ -22,6 +22,24 @@ Join us in our gitter channel: https://gitter.im/spring-cloud-app-broker/communi
 
 === Compatibility
 
+==== 1.3.x
+
+* https://projects.spring.io/spring-framework/[Spring Framework] 5.3.x
+* https://projects.spring.io/spring-boot/[Spring Boot] 2.4.x
+* https://github.com/cloudfoundry/cf-java-client/[Cloud Foundry Java Client] 5.x
+* https://github.com/reactor/[Reactor] 3.4.x
+* https://spring.io/projects/spring-cloud-open-service-broker/[Spring Cloud Open Service Broker] 3.4.x
```
royclarkson, line 31: Let's keep on SCOSB 3.3.x since it's built on Boot 2.4 and already GA.

README.adoc
```diff
@@ -22,6 +22,24 @@ Join us in our gitter channel: https://gitter.im/spring-cloud-app-broker/communi
...
+* https://github.com/openservicebrokerapi/servicebroker/tree/v2.15/[Open Service Broker API] 2.16
```
royclarkson, line 32: Same as above, let's stick on 2.15 via SCOSB.

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
```diff
@@ -114,7 +114,7 @@
 	/**
 	 * The buildpack to use for deploying the application.
 	 */
-	private String buildpack;
+	private String buildpack = "java_buildpack_offline";
```
royclarkson, line 117: seems reasonable to me. @spikymonkey any concerns about this?

royclarkson (changes requested):
added a couple minor changes in compatibility matrix

README.adoc
Albertoimpl: Good point. Changed!

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
spikymonkey: What's the context around the need for the change? I would probably favour keeping things explicit rather than defaulting - we can't guarantee which buildpacks are going to be available in the target env, and I've certainly seen envs where `java_buildpack_offline` doesn't exist.

royclarkson: CFJC now requires a build pack name in some API calls.

Albertoimpl: The context is that buildpacks are not optional anymore, they have to be explicitly specified.
For example https://github.com/spring-cloud/spring-cloud-app-broker/pull/409/files#diff-6ed7594932ade9fa992e723b4f269c0b61cc4127a8dc40af23c9ab1168cc6e53R54
With that default we just make it work in most of the PCF environments, but it's true that `java_buildpack_offline` is not available in other non PCF environments.
We have to document it regardless of the default that `buildpack` is not optional anymore, regardless if we default it or not.

spikymonkey: Wait what? 😅

spikymonkey: That's inconvenient

README.adoc
royclarkson: 👍

royclarkson (approved):
Looks good!

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryDeploymentProperties.java
spikymonkey: I can't really tell where the mandatory buildpack has come in from this, but it doesn't make much sense to me.  User should totally be able to throw some bits at the API and have the available buildpacks determine which if any can convert those bits into a package. Any chance we can use the buildpack returned from the build creation step in further API calls?

Albertoimpl: To be more concise, the buildpacks are now mandatory when doing a push with manifest, which was not the case before and it's what we use to do most of the create operations. 
We don't have much information about buildpacks before that.

spikymonkey: Can it just be set to an empty array (i.e. autodetect buildpack)? 

Screenshot 2021-02-18 at 5 30 15 pm

Albertoimpl: Definitely much better! I will change it to that. Thanks!

Albertoimpl: Updated it to an empty String to construct the manifest as an empty buildpacks array.

spikymonkey (approved):
LGTM @Albertoimpl - thanks!

PR #410, opened 2021-02-23
Updating CFJC to 5.1

Albertoimpl:
Updating compatibility matrix with SCOSB and OSBAPI versions

Updating default buildpack to empty String. Nulls are not allowed by operations and the only way to enable autodiscovery is by passing and empty list, passing an empty String will trigget that behaviour. Updating tests to document the default buildpack


spikymonkey (approved):

Albertoimpl:
Thanks @spikymonkey!

PR #415, opened 2021-03-11
Use status field to check deployment status instead of state

alek-sys:
State field was deprecated in CAPI and starting from CFJC 5.3 marked as nullable, since it won't be sent by CAPI versions 3.91+.

App Broker should use status field instead.


Albertoimpl (approved):
Looks great! Thanks!

PR #416, opened 2021-03-11
Only recreate client if the existing client has different authorities

LittleBaiBai:
Previously we recreate client before each test. When multiple tests are running concurrently, if a token is issued for test A but the client was deleted during test B, then the token used for test A would be an 'invalid token' and fail the test. Another error scenario is when test A and test B both trigger client creation at the same time, one of them would error out due to `client already exists`. This fix should avoid most of the risk in terms of client creation


royclarkson (approved):
Looks ok 🤷‍♂️   I saw the passing test runs, so go for it!

PR #419, opened 2021-04-15
Update ci-image source to harbor

dyroberts:


royclarkson (approved):
Are these consistent with other projects? Otherwise all looks good. 👍

PR #457, opened 2021-09-06
Fix dependencies for the 1.1.x branch

spikymonkey:
I think we picked up some unintended commits in this branch since the
last 1.1.x release tag and accidentally moved to Spring Boot 2.3.
Rolling back to the latest available versions in the correct release
series:

* CF Java Client 4.16.0.RELEASE
* Open Service Broker 3.1.2.RELEASE
* Spring Boot 2.2.13.RELEASE


Albertoimpl (approved):
Looks good! It also looks like will keep the compatibility with OSBAPI  2.25 even with these versions.

PR #465, opened 2021-09-10
Update version after release

abelsromero:


abelsromero:
NOTE: the commit for the Dockerfile update was not included because we filter that file in put git-repo during build task. Is fine.

PR #474, opened 2021-09-23
Scaling  vm types 1.1.x

Albertoimpl:
By default, cf-deployment uses minimal vm type, scaling the main components up vertically
to use large instead.


spikymonkey (approved):
Thanks for doing this @Albertoimpl! 👍

PR #504, opened 2022-01-05
Updating concourse vars

Albertoimpl:


ci/tasks/build-project.yml
```diff
@@ -15,8 +15,7 @@ caches:
   - path: maven
   - path: gradle
 run:
-  path: ci/scripts/build-project.sh
-  dir: git-repo
+  path: git-repo/ci/scripts/build-project.sh
```
spikymonkey, line 18: Another solution could be to pass `DISTRIBUTION_REPOSITORY_OUTPUT` in the `params` rather than try to compute it from `pwd` in the script. We use that pattern in most of our other projects for the input / output paths... wdyt?

spikymonkey (approved):

PR #560, opened 2022-02-08
Updating release docs to include extra checks

Albertoimpl:
And to add how to recover from some common problems

### DIFF
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -116,7 +116,7 @@ Downloading org/springframework/cloud/spring-cloud-app-broker-security-credhub/1
 Downloading org/springframework/cloud/spring-cloud-app-broker-security-credhub/1.4.0 from libs-staging-local
 ----
 
-We can go and manually verify that it contains the right values.
+We can go and manually verify that it contains the right values by, for example, downloading the `.pom` file in https://repo.spring.io/ui/builds/spring-cloud-app-broker?buildRepo=artifactory-build-info
 
 Once we are happy with the staged release, we can go and promote that build to be released by manually executing the `promote-release` job.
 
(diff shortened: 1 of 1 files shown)


RELEASING.adoc
```diff
@@ -134,7 +134,22 @@ The task output should be something similar to:
...
-Now, we should be able to find the commits bumping the versions, that we should manually merge in order to maintain a linear history https://github.com/spring-cloud/spring-cloud-app-broker/commit/e24b81505b79b59b1cff0206d8166227a7353400
+We can check that it worked in https://repo.spring.io/ui/artifactSearchResults?name=spring-cloud-app-broker&type=artifacts
+
+Now, we should be able to find the commits bumping the versions, that we should manually merge in order to maintain a linear history https://github.com/spring-cloud/spring-cloud-app-broker/commits/spring-builds/staging
+
+It can be done by creating a Pull Request cherry-picking the two commits in order:
+
+----
+git checkout -b next-release-version
+git cherry-pick 1c9e0b9709674674f51618865c0057b18e77f754
```
spikymonkey, line 145: Is this necessary or can we just create a pull request directly from the `staging` branch?

spikymonkey (commented):
Thanks for taking the time to improve this :)

RELEASING.adoc
Albertoimpl: Didn't try that option, to be honest. I always try to use the CLI and this should be equivalent if I am not mistaken.

spikymonkey: I think it's equivalent yes, but what do we gain by doing it when we still have to create a PR anyway? The way I'm thinking is, can we remove potential for human error by not having to copy/paste SHAs?

Albertoimpl: Sure. Just updated it. Thanks!

spikymonkey (approved):

PR #584, opened 2022-03-31
Upgrading boot to 2.5.12

Albertoimpl:


build.gradle
```diff
@@ -26,7 +26,7 @@ buildscript {
 		immutablesVersion = "2.9.0"
 		openServiceBrokerVersion = "3.4.1"
 		pmdVersion = "6.29.0"
-		springBootVersion = "2.5.11"
+		springBootVersion = "2.5.12"
 		springCredhubVersion = "2.2.0"
 		springFrameworkVersion = "5.3.14"
```
royclarkson, line 31: this needs to be updated too

royclarkson (changes requested):
spring framework version also needs to be bumped. I might have missed that on the other PRs.

PR #641, opened 2022-10-13
Upgrade to Gradle 7.5.1

spikymonkey:
Also default to running tasks in parallel


royclarkson (approved):

spikymonkey:
Thanks @royclarkson 👍

PR #651, opened 2022-10-21
Remove enforcedPlatform dependency management

spikymonkey:
Allowing consumer projects to apply their own version management


Albertoimpl (approved):
👏🏽

PR #697, opened 2023-01-19
Use spring-boot-backends

Buzzardo:
to define the look and feel of the HTML output.

Part of Spring's continuing doc modernization effort.


Albertoimpl (approved):
Thank you very much!

PR #705, opened 2023-01-31
Adding buildpacks to deployer properties

Albertoimpl:
In order to enable multiple buildpacks to be added to the backing application manifest


abelsromero (approved):
Thanks for the explanation!!

abelsromero (approved):

PR #713, opened 2023-02-08
Fixing space parallelization

### LINKED ISSUE #701: Service instances race condition
When creating more than one service instance at the same time, there might be a race condition where both threads will try to create the backing space and it will make one of the two fails.
Instead of failing, we should see if the problem is that the space already exists and continue with the service creation.
If this causes problems, we can always fallback to using `concatMap`, although this will be a bit less performant.

Albertoimpl:
[Fixes #701]

### DIFF
--- a/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
+++ b/spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
@@ -767,6 +767,7 @@ public class CloudFoundryAppDeployer implements AppDeployer, ResourceLoaderAware
 						})
 						.doOnError(e -> LOG.error(String.format("Error creating space. spaceName=%s, " +
 							ERROR_LOG_TEMPLATE, spaceName, e.getMessage()), e))
+						.onErrorResume(e -> Mono.empty())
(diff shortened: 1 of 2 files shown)


spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -767,6 +767,7 @@ private Mono<String> createSpace(String spaceName) {
 						})
 						.doOnError(e -> LOG.error(String.format("Error creating space. spaceName=%s, " +
 							ERROR_LOG_TEMPLATE, spaceName, e.getMessage()), e))
+						.onErrorResume(e -> Mono.empty())
```
abelsromero, line 770: Wound't this make the scenario to fail silently?

abelsromero (approved):

spring-cloud-app-broker-deployer-cloudfoundry/src/main/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployer.java
Albertoimpl: The opposite, when the two threads try to create the same space it typically fails, now we can assume that when the space exists, we can carry on with the flow.

abelsromero (approved):

PR #746, opened 2023-04-17
Removing credentials provider, logging and security module

Albertoimpl:


royclarkson (approved):
That's a lot of stuff to delete! 😅

Albertoimpl:
It doesn't get more productive than this 😄 
Thanks for the review!

PR #782, opened 2023-05-25
(v2.0.x) Delete dependabot from 2.0.x branch

abelsromero:
v2.0.x branch dependabot allows for minor bump, which caused this PR https://github.com/spring-cloud/spring-cloud-app-broker/pull/779.

This PR aligns the configuration already present in main branch to v2.0.x. I have doubts whether having the config in only the main branch is enough, there are recent reports contradicting it https://github.com/spring-cloud/spring-cloud-app-broker/pull/779


abelsromero:
> Shouldn't we just remove this file from all branches other than `main`?

I am happy to do it, we'll see if everything works fine 🚀

Issue #783, opened 2023-05-25
Add target branch to dependabot PR title


abelsromero:
This helps to managing the same bump for multiple branches or identifying bumps that should not be accepted.
Only requires adding this configuration

```
    commit-message:
      prefix: "(v2.0.x)"
```

patibachleda:
Thank you for making this an issue as it gave me a chance to look into GitHub PR flow from an open source project!

patibachleda:
I made separate pull requests for each branch adding the commit message prefixes for each, I hope that is the correct way of doing this. I wasn't sure if I should also do the commit messages for 1.2.x, 1.1.x, and 1.0.x.

abelsromero:
Closed with https://github.com/spring-cloud/spring-cloud-app-broker/pull/792

PR #784, opened 2023-05-26
(main) Add target branch to dependabot PR title

patibachleda:
Addressing [Issue#783](https://github.com/spring-cloud/spring-cloud-app-broker/issues/783)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     schedule:
       interval: daily
     commit-message:
-      prefix: "(v2.0.x)"
+      prefix: "(2.1.x)"
     milestone: 29
     labels:
       - dependencies
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,8 @@ updates:
     target-branch: "2.0.x"
     schedule:
       interval: daily
+    commit-message:
+      prefix: "(2.0.x)"
     milestone: 26
     labels:
       - dependencies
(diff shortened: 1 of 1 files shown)


abelsromero:
Looking at it now, I realized my example in the issue was not correct :sweat: I took it from a repo where we prefix all branches with `v`, here we should just use `2.1.x`, `2.0.x`, `1.6.x`, `1.5.x`, and `1.4.x`.
Would you mind fixing this and adding the missing `commit-message:` (in short, all have the same as branch except for main)?

| target-branch  | prefix |
| ----------- | ----------- |
| main      | 2.1.x       |
| 2.0.x   | 2.0.x        |
| 1.6.x   | 1.5.x        |
| 1.5.x   | 1.5.x        |
| 1.4.x   | 1.4.x        |


Btw, feel free to ask any question, and thanks a lot for contributing.

abelsromero:
My mistake again, sorry. All configurations should be kept in the main branch, see I already created PRs to delete dependabot in the other branches.
You can make all the changes in this same PR in a single file, adding the the config to each block.

image

Apologies again, you got us in between some dependabot cleanup :sweat_smile:

.github/dependabot.yml
```diff
@@ -34,6 +36,8 @@ updates:
     target-branch: "1.6.x"
     schedule:
       interval: daily
+    commit-message:
+      prefix: "(1.5.x)"
```
abelsromero, line 40: I think you mean `(1.6.x)`  here?

abelsromero (changes requested):

.github/dependabot.yml
abelsromero: The rest is OK :ok_hand: thanks

abelsromero (approved):
Thanks a lot!

PR #798, opened 2023-06-25
(1.6.x) Remove Spring Boot duplication in configuration

abelsromero:
This allows removing `springBootVersion` so that the actual version can be set only once in the boot plugin in 'settings.gradle'. This requires adding the boot plugin to all modules, modules that do not use use 'apply false' to only add the plugin libraries to the build runtime.

This will fix the inconsistency of dependabot updates, because we are getting 2 PRs for each Boot bump, instead of one with the updated version in all places.


spikymonkey (approved):
Nice improvement, thanks @abelsromero!

PR #799, opened 2023-06-25
 (1.5.x) Remove Spring Boot duplication in configuration

abelsromero:
This allows removing 'springBootVersion' so that the actual version can be set only once in the boot plugin in 'settings.gradle'. This requires adding the boot plugin to all modules, modules that do not use use 'apply false' to only add the plugin libraries to the build runtime.

This will fix the inconsistency of dependabot updates, because we are getting 2 PRs for each Boot bump, instead of one with the updated version in all places.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,6 @@ buildscript {
 		immutablesVersion = "2.9.3"
 		openServiceBrokerVersion = "3.5.4"
 		pmdVersion = "6.29.0"
-		springBootVersion = "2.6.13"
 		springCredhubVersion = "2.2.0"
 		springFrameworkVersion = "5.3.22"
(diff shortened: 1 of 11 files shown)


spring-cloud-starter-app-broker/build.gradle
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+import org.springframework.boot.gradle.plugin.SpringBootPlugin
+
+plugins {
+	id 'org.springframework.boot' apply false
```
spikymonkey, line 20: Is this needed here?

abelsromero: No, good catch, thanks!

spikymonkey (approved):

PR #842, opened 2023-09-05
Removing CredHub dependency from buildscript

Albertoimpl:
Since we deleted the credhub-security module as part of https://github.com/spring-cloud/spring-cloud-app-broker/commit/135e3496d112aff71e06a0a6a46777bac3f7e7c7, we don't need this dependency in the buildscript anymore


abelsromero (approved):
Thanks!

PR #846, opened 2023-09-06
1.6.x Updating wiremock to a java8 compatible version

Albertoimpl:


abelsromero (approved):
Approving to run CI

Albertoimpl:
Dependabot only gets triggered from what is on `dependabot.yml` in `main`  if I remember correctly.
I am adding these lines in main too:
```
      - dependency-name: "com.github.tomakehurst:wiremock-jre8-standalone"
        update-types:
          - "version-update:semver-minor"
          - "version-update:semver-patch"
```

PR #848, opened 2023-09-06
Renaming wiremock artifact ID to the latest

Albertoimpl:


abelsromero (dismissed):
Same as the other, with the new group id

abelsromero (approved):

Issue #831, opened 2023-08-30
Support 'stack' as a configurable app parameter


spikymonkey:
Allow users to specify the stack (e.g. `cflinuxfs3`, `cflinuxfs4`) that should be used to run their application. The user may want to specify an ordered list of stacks, e.g. `cflinuxfs4`, or fall back to `cflinuxfs3` if not available.

PR #856, opened 2023-09-14
Adding stack support

Albertoimpl:
[Finishes #831]

### DIFF
--- a/spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
+++ b/spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
@@ -208,6 +208,7 @@ class CloudFoundryAppDeployerTest {
 			.property(CloudFoundryDeploymentProperties.HEALTHCHECK_PROPERTY_KEY, "http")
 			.property(CloudFoundryDeploymentProperties.HEALTHCHECK_HTTP_ENDPOINT_PROPERTY_KEY, "/healthcheck")
 			.property(CloudFoundryDeploymentProperties.BUILDPACKS_PROPERTY_KEY, "buildpack1,buildpack2")
+			.property(CloudFoundryDeploymentProperties.STACK_PROPERTY_KEY, "customstack")
 			.property(CloudFoundryDeploymentProperties.DOMAINS_PROPERTY, "domain1,domain2")
 			.property(DeploymentProperties.HOST_PROPERTY_KEY, "host")
 			.property(CloudFoundryDeploymentProperties.NO_ROUTE_PROPERTY, "true")
@@ -226,6 +227,7 @@ class CloudFoundryAppDeployerTest {
 			.healthCheckType(ApplicationHealthCheck.HTTP)
 			.healthCheckHttpEndpoint("/healthcheck")
 			.buildpacks("buildpack1", "buildpack2")
+			.stack("customstack")
 			.domains("domain2", "domain1") // domains is a list so order matters
 			.host("host")
 			.noRoute(true)
@@ -295,6 +297,7 @@ class CloudFoundryAppDeployerTest {
 			.property(CloudFoundryDeploymentProperties.DOMAINS_PROPERTY, "domain2")
 			.property(DeploymentProperties.HOST_PROPERTY_KEY, "host2")
 			.property(CloudFoundryDeploymentProperties.NO_ROUTE_PROPERTY, "true")
+			.property(CloudFoundryDeploymentProperties.STACK_PROPERTY_KEY, "customstack")
 			.build();
 
 		StepVerifier.create(appDeployer.deploy(request))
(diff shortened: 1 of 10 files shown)


spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
```diff
@@ -208,6 +208,7 @@ void deployAppWithPropertiesInRequest() {
 			.property(CloudFoundryDeploymentProperties.HEALTHCHECK_PROPERTY_KEY, "http")
 			.property(CloudFoundryDeploymentProperties.HEALTHCHECK_HTTP_ENDPOINT_PROPERTY_KEY, "/healthcheck")
 			.property(CloudFoundryDeploymentProperties.BUILDPACKS_PROPERTY_KEY, "buildpack1,buildpack2")
+			.property(CloudFoundryDeploymentProperties.STACK_PROPERTY_KEY, "customstack")
```
spikymonkey, line 211: I think this needs to be a list, like we have for buildpacks, so allow for fallbacks. I.e. 'give me `mycustomfunkystack` if available, otherwise fallback to cflinuxfs4, then cflinuxfs3'

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/WiremockComponentTest.java
```diff
@@ -47,7 +47,8 @@
 		"spring.cloud.appbroker.deployer.cloudfoundry.default-org=test",
 		"spring.cloud.appbroker.deployer.cloudfoundry.default-space=development",
 		"spring.cloud.appbroker.deployer.cloudfoundry.secure=false",
-		"spring.cloud.appbroker.deployer.cloudfoundry.properties.buildpack=example-buildpack"
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.buildpack=example-buildpack",
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.buildpack=cflinuxfs4"
```
spikymonkey, line 51: Was this meant to be stack, not buildpack?

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/fixtures/CloudControllerStubFixture.java
```diff
@@ -183,6 +183,12 @@ public void stubAppDoesNotExist(final String appName) {
 				.withBody(cc("empty-query-results"))));
 	}
 
+	public void stubListStacks() {
+		stubFor(get(urlPathEqualTo("/v2/stacks"))
```
spikymonkey, line 187: Can we avoid using the dead v2 API for new code?

Albertoimpl: We are not explicitly calling stacks, it is the CF Java client during the push that validates the stacks and it uses that API call. Not much we can do without updating CFJC.

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/WiremockComponentTest.java
Albertoimpl: Oh, I should have that deleted. I may add the (right) property there if I can't create a good AT.
Thanks!

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/fixtures/CloudControllerStubFixture.java
spikymonkey: Oh right... CFJC's `CloudFoundryClient` has `StacksV3 stacksV3()`, but maybe the Operations API is still going via V2?

spring-cloud-app-broker-deployer-cloudfoundry/src/test/java/org/springframework/cloud/appbroker/deployer/cloudfoundry/CloudFoundryAppDeployerTest.java
spikymonkey: Ah okay thanks... I guess there's a question of whether we support fallback at the appbroker or service broker level then?

Albertoimpl: do we have to support fallback? Since we are not defaulting to anything in app-broker it will default already to what the platform states.
If the service broker implementer decides to overwrite it with a different value, then, that must exist.
Happy with including a separate story to talk about fallbacks, but I do not see it as a must yet.

spring-cloud-app-broker-integration-tests/src/test/java/org/springframework/cloud/appbroker/integration/fixtures/CloudControllerStubFixture.java
Albertoimpl: Exactly that.
Adding a note to review the v2 deprecation policy to see if we should start thinking about an OperationsV3.

spikymonkey (approved):
Great, thanks @Albertoimpl!

PR #868, opened 2023-09-28
Improving CF deployer documentation

Albertoimpl:
Adding stack as a suported property

### DIFF
--- a/spring-cloud-app-broker-docs/src/docs/asciidoc/service-instances.adoc
+++ b/spring-cloud-app-broker-docs/src/docs/asciidoc/service-instances.adoc
@@ -43,6 +43,26 @@ The following table lists properties that can be set for all or for deployments:
 |Description
 |Default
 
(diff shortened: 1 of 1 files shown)


spikymonkey (dismissed):
Thanks @Albertoimpl! Should we take the opportunity to sort the rows in this table? It feels a bit random at the moment.

Albertoimpl:
I like the idea, should we introduce sub-groups or go with alphabetical order @spikymonkey?

spikymonkey:
> I like the idea, should we introduce sub-groups or go with alphabetical order @spikymonkey?

Maybe alphabetical for now, since it's a fairly short list? We can always add sub groups later if we want to.

Albertoimpl:
Done, thanks @spikymonkey!

spikymonkey (approved):
Cool, thanks @Albertoimpl!

PR #869, opened 2023-10-02
Moving pipelines to shepherd

Albertoimpl:


ci/scripts/acceptance-tests.sh
```diff
@@ -67,27 +62,26 @@ prepare_cf() {
...
+  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_API_HOST="${CF_API_HOST}"
+  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_API_PORT=443
+  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_USERNAME="${CF_USERNAME}"
+  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_PASSWORD="${CF_PASSWORD}"
   export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_CLIENT_ID="${CLIENT_ID}"
   export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_CLIENT_SECRET="${CLIENT_SECRET}"
   export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_DEFAULT_ORG="${DEFAULT_ORG}"
   export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_DEFAULT_SPACE="${DEFAULT_SPACE}"
-  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_SKIP_SSL_VALIDATION="${SKIP_SSL_VALIDATION}"
+  export SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_SKIP_SSL_VALIDATION=true
```
spikymonkey, line 73: Really? Toolsmiths envs had valid certs... Do the Shepherd ones not? Feels like that is something we should raise if so. And is there a reason to remove the variable?

ci/scripts/acceptance-tests.sh
```diff
@@ -67,27 +62,26 @@ prepare_cf() {
...
   export TESTS_BROKERAPPPATH=build/libs/spring-cloud-app-broker-acceptance-tests.jar
   ./gradlew -PacceptanceTests \
-    -PonlyShowStandardStreamsOnTestFailure="${ONLY_SHOW_STANDARD_STREAMS_ON_TEST_FAILURE}" \
+    -PonlyShowStandardStreamsOnTestFailure=true \
```
spikymonkey, line 76: Why this change?

scripts/acceptance-tests-ide.sh
```diff
@@ -2,46 +2,84 @@
...
+  eval "$(bbl print-env --metadata-file "${METADATA_FILE}")"
 
-  API_HOST="$(jq -r .cf.api_url <"$TOOLSMITH_ENV_INPUT/metadata")"
-  PASSWORD="$(credhub get -n "/bosh-${env_name}/cf/cf_admin_password" -q)"
+  CF_API_HOST="$(jq -r .cf.api_url <"${METADATA_FILE}")"
+  CF_PASSWORD="$(credhub get -n "/bosh-${env_name}/cf/cf_admin_password" -q)"
   CLIENT_SECRET="$(credhub get -n "/bosh-${env_name}/cf/uaa_admin_client_secret" -q)"
 }
 
+prepare_cf_deployment() {
```
spikymonkey, line 30: Thanks! Not sure how I feel about this since it can take a very long time to run. The original intent of the script was to just dump out the vars to paste into IJ. I wonder if this should be broken out into a separate `prepare-cf-environment.sh` or something like that?

ci/scripts/acceptance-tests.sh
Albertoimpl: It was always `true` in the task file. I was trying to remove as many inputs as possible so we only need the metadata file.

ci/scripts/acceptance-tests.sh
Albertoimpl: Oh, I thought It was always hardcoded to `true`, that's why I defaulted it to that. Let me check if we can default it to `false`.
I have been trying to remove inputs to make the script easier to run locally.

scripts/acceptance-tests-ide.sh
Albertoimpl: I considered a few things:
- Removing it, but it didn't scale for our ATs
- Creating a file with the replace ops.yml but we need to get the deployment and every time you want to run the ATs, you have to scale the environment up anyway. The second time it just moves forward so it doesn't take that long to leave it as part of the same script.
- Preparing a different pool with everything scaled up already, but I will have to do more research on that.

I decided to leave it as it was since it has been helpful for me running the tests locally.

ci/scripts/acceptance-tests.sh
Albertoimpl: Sadly, just tested it, It does not have valid certs.

spikymonkey: I should have looked more closely, I see the var is `:?must be set` now. I think it's fine.

ci/pr-pipeline.yml
```diff
@@ -95,10 +96,14 @@ resources:
...
-      hostname: environments.toolsmiths.cf-app.com
-      pool_name: cf-deployment
+      url: https://v2.shepherd.run
+      service-account-key: ((shepherd-service-account-key))
+      lease:
+        namespace: spring-cloud-app-broker
+        pool:
+          namespace: official
+          name: cfd
+      compatibility-mode: environments-app
```
spikymonkey, line 109: Please could you fix the trailing newline? We should probably have this set in `.editorconfig` if we don't already.

scripts/acceptance-tests-ide.sh
```diff
@@ -2,46 +2,84 @@
...
-
-declare API_HOST
-readonly API_PORT=443
-readonly USERNAME="admin"
-declare PASSWORD
-readonly CLIENT_ID="admin"
+readonly METADATA_FILE="${METADATA_FILE:?must be set}"
+
+declare CF_API_HOST
+declare CF_USERNAME="admin"
```
spikymonkey, line 8: Can these vars with fixed values be made `readonly`?

scripts/acceptance-tests-ide.sh
```diff
@@ -2,46 +2,84 @@
...
+  cf create-space "$DEFAULT_SPACE" -o "$test_instances_org"
+}
+
 print_values() {
-cat <<EOF
-SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_API_HOST=${API_HOST}
-SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_API_PORT=${API_PORT}
-SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_USERNAME=${USERNAME}
-SPRING_CLOUD_APPBROKER_ACCEPTANCETEST_CLOUDFOUNDRY_PASSWORD=${PASSWORD}
+  cat <<EOF
```
spikymonkey, line 65: Whitespace

spikymonkey (changes requested):
Thanks Alberto! Some very minor things, but really looking forward to seeing this working!

ci/pr-pipeline.yml
Albertoimpl: Updated the `.editorconfig` it was not being picked up and had some missing things for the non-java files. Thanks!

scripts/acceptance-tests-ide.sh
Albertoimpl: Isn't this good?
```
print_values() {
  echo "We have this"
}

print_values() {
echo "We had this"
} 
 ```
 ```

File: scripts/acceptance-tests-ide.sh (hunk 5 above)
Comment (Albertoimpl) [reply scripts/acceptance-tests-ide.sh:8]: Ah yes, thanks!

File: ci/scripts/acceptance-tests.sh [hunk 7]
```diff
@@ -2,37 +2,38 @@
...
+declare CF_PASSWORD
+
 readonly CLIENT_ID="admin"
 declare CLIENT_SECRET
 
+readonly DEFAULT_ORG="test"
+readonly DEFAULT_SPACE="development"
+
+DEPLOYMENT_DIRECTORY="$(mktemp -d)"
+readonly DEPLOYMENT_DIRECTORY
```
Albertoimpl, line 18: TIL this is the expected way to assign and declare a `readonly`: https://www.shellcheck.net/wiki/SC2155

spikymonkey: Shellcheck ftw!

scripts/acceptance-tests-ide.sh
spikymonkey: Sorry, my bad! I was thrown by the diff but you're totally right.

spikymonkey (approved):
Thanks @Albertoimpl!

Albertoimpl:
Thanks a lot for the review @spikymonkey!

PR #878, opened 2023-10-05
Branch for 2.1.x series

spikymonkey:


Albertoimpl:
To make it consistent with how we did it in the past, can you also update the dependabot and the main readme? https://github.com/spring-cloud/spring-cloud-app-broker/commit/7303c938c9eb956dc567130bc5083181f39c0a3b

spikymonkey:
Ah yes of course! I've updated the dependabot config, and created a new section in the README, but since we've just branched I think we'll need to leave updating the key dependency versions in there until we know more.

Albertoimpl (approved):
Thanks!

spikymonkey:
Thanks @Albertoimpl!

PR #965, opened 2024-03-01
Only allow patch bumps to Spring Boot

spikymonkey:
`dependency-name` was slightly wrong, due to Boot version being managed via the plugin, rather than a normal dependency.


Albertoimpl (approved):
Thanks!

spikymonkey:
Thanks @Albertoimpl!

PR #973, opened 2024-03-06
Removing stack assertion in ATs

Albertoimpl:
Moving all references from cflinuxfs3 to cflinuxfs4.
Most of the cf deployments don't have any other stack aside from cflinuxfs4, moved the assertions to check on 4.

### DIFF
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceAcceptanceTest.java
@@ -70,7 +70,7 @@ class CreateInstanceAcceptanceTest extends CloudFoundryAcceptanceTest {
 		"spring.cloud.appbroker.services[0].apps[1].name=" + APP_CREATE_2,
 		"spring.cloud.appbroker.services[0].apps[1].path=" + BACKING_APP_PATH,
 
-		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=cflinuxfs3"
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=cflinuxfs4"
 	})
 	void deployAppsOnCreateService() {
 		// when a service instance is created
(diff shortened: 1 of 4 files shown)


spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceAcceptanceTest.java
```diff
@@ -69,8 +68,6 @@ protected String backingServiceName() {
 
 		"spring.cloud.appbroker.services[0].apps[1].name=" + APP_CREATE_2,
 		"spring.cloud.appbroker.services[0].apps[1].path=" + BACKING_APP_PATH,
-
-		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=cflinuxfs3"
```
Albertoimpl, line 73: Since `cflinuxfs3` is not present anymore I moved this assertion to a new test.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CreateInstanceWithStackAcceptanceTest.java
```diff
@@ -0,0 +1,82 @@
...
+		"spring.cloud.appbroker.services[0].apps[0].path=" + BACKING_APP_PATH,
+
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=custom-stack"
+	})
+	void deployAppsOnCreateService() {
+		// when a service instance is created
+		createServiceInstance(SI_NAME);
+
+		// and stack is updated when specified
+		Optional<ApplicationDetail> application1Detail = getApplicationDetail(APP_CREATE_1);
```
Albertoimpl, line 69: Not that the application won't start successfully since this stack does not exist, but we only care that the stack was changed.

Albertoimpl:
Since we have an integration test that covers that the stack is being passed to the platform, I don't think we are loosing much from not checking that the stack changes.

spikymonkey (approved):
Thanks @Albertoimpl!

PR #986, opened 2024-03-12
Qualifying ConnectionContext 

Albertoimpl:
To avoid conflicts when using app-broker with other contexts

Related to https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x#parameter-name-retention


spikymonkey (approved):
Nice!

PR #988, opened 2024-03-12
Make PR manager use the default branch

Albertoimpl:


scripts/set-pipelines.sh
```diff
@@ -27,13 +27,15 @@ set_branch_pipeline() {
 
 set_pr_manager_pipeline() {
 	local -r pipeline_name="app-broker-pull-requests${PIPELINE_NAME_SUFFIX:+"-$PIPELINE_NAME_SUFFIX"}"
+  local -r branch="2.2.x"
```
spikymonkey: Indentation looks off

spikymonkey (changes requested):
Thanks @Albertoimpl , just a small whitespace issue

scripts/set-pipelines.sh
Albertoimpl: oh, thanks! Fixed

spikymonkey (approved):
Thanks!

PR #1064, opened 2024-07-04
(2.3.x): Upgrading spring boot to 3.3.1

Albertoimpl:


kvmw (approved):

spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/state/InMemoryServiceInstanceStateRepositoryTest.java
```diff
@@ -44,7 +44,7 @@ void saveAndGet() {
 				assertThat(serviceInstanceState.getOperationState()).isEqualTo(OperationState.IN_PROGRESS);
 				assertThat(serviceInstanceState.getDescription()).isEqualTo("bar");
 				assertThat(serviceInstanceState.getLastUpdated())
-					.isEqualToIgnoringSeconds(new Date());
+					.isCloseTo(new Date(), 1000L);
```
spikymonkey, line 47: `isCloseTo` is definitely the right choice here, but are we sure 1 sec is enough to avoid flakes?

spikymonkey (approved):

spring-cloud-app-broker-core/src/test/java/org/springframework/cloud/appbroker/state/InMemoryServiceInstanceStateRepositoryTest.java
Albertoimpl: I tested every method locally with `@RepeatedTest(10000)` and nothing failed.
Just did it again with 100.000 repetitions and they were all still happy.

Issue #995, opened 2024-03-20
Restore service-instance log proxying by using recent-log endpoints on the CF log-cache API


emalm:
The commit at https://github.com/spring-cloud/spring-cloud-app-broker/commit/09d4736ceb4075e0a254440d7f2c037409bb3dd7 appears to have removed the ability of this broker to proxy logs for the CF app instances that constitute the service instances that it manages. This removal was prompted by the Traffic Controller removing the recent-logs endpoint, as part of CF's transition away from the Loggregator system in favor of its [new "shared-nothing" architecture](https://docs.vmware.com/en/VMware-Tanzu-Application-Service/5.0/tas-for-vms/loggregator-architecture.html#sharednothing-architecture-2), but there are equivalent HTTP endpoints on the log-cache component that allow consumers such as the CF CLI and Apps Manager to get recent logs by app ID (or source ID, more generally). The log-cache component is expected to remain part of the CF/TAS logging architecture even as the rest of the Loggregator system is deprecated and removed over time, so its endpoints should be stable indefinitely.

More information about these log-cache endpoints is available here:
* Documentation for the log-cache HTTP API: https://github.com/cloudfoundry/log-cache-release/tree/main/src#get-apiv1readsource-id
* https://github.com/pivotal-cf/spring-cloud-gateway-service/issues/2058#issuecomment-1503308547 has an overview of the API endpoints and responses.
* https://github.com/cloudfoundry/go-log-cache is a Golang client library for the log-cache API, and is used in the [log-cache cf CLI plugin](https://github.com/cloudfoundry/log-cache-cli).

cc @mkocher @acrmp @ctlong @rroberts2222

Albertoimpl:
Thanks a lot for all the information. 
Fixed by [ca6f4278e4cfcf7a5e75adec9efdaa89151e6fc4](https://github.com/spring-cloud/spring-cloud-app-broker/commit/ca6f4278e4cfcf7a5e75adec9efdaa89151e6fc4)

PR #1065, opened 2024-07-04
(2.3.x): Adding logging module using LogCacheClient

Albertoimpl:
[Finishes #995]

### DIFF
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
@@ -532,12 +532,6 @@ abstract class CloudFoundryAcceptanceTest {
 			future.cancel(true);
 			executor.shutdownNow();
 			process.destroyForcibly();
-			try {
-				process.waitFor(5, TimeUnit.SECONDS);
-			}
-			catch (InterruptedException e) {
-				LOG.warn("Interrupted while waiting for process to terminate", e);
-			}
 		}
 		return outputBuilder.toString();
 	}
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,9 +16,14 @@
 
 package org.springframework.cloud.appbroker.acceptance;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.net.URI;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -29,6 +34,12 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import com.jayway.jsonpath.Configuration;
 import com.jayway.jsonpath.DocumentContext;
@@ -64,6 +75,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryClientConfiguration;
+import org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryProperties;
 import org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFoundryService;
 import org.springframework.cloud.appbroker.acceptance.fixtures.cf.UserCloudFoundryService;
 import org.springframework.cloud.appbroker.acceptance.fixtures.uaa.UaaService;
@@ -91,6 +103,7 @@ import static org.springframework.cloud.appbroker.acceptance.fixtures.cf.CloudFo
 @ExtendWith(SpringExtension.class)
 @ExtendWith(BrokerPropertiesParameterResolver.class)
 @EnableConfigurationProperties(AcceptanceTestProperties.class)
+@SuppressWarnings("PMD.GodClass")
 abstract class CloudFoundryAcceptanceTest {
 
 	private static final Logger LOG = LoggerFactory.getLogger(CloudFoundryAcceptanceTest.class);
@@ -114,10 +127,15 @@ abstract class CloudFoundryAcceptanceTest {
 	protected UserCloudFoundryService userCloudFoundryService;
 
 	@Autowired
+	private CloudFoundryProperties cloudFoundryProperties;
+
+	@Autowired
 	private UaaService uaaService;
 
 	@Autowired
 	private AcceptanceTestProperties acceptanceTestProperties;
+
+	private String cfHome;
 
 	private final WebClient webClient = getSslIgnoringWebClient();
 
@@ -144,6 +162,7 @@ abstract class CloudFoundryAcceptanceTest {
 		List<String> appBrokerProperties = getAppBrokerProperties(brokerProperties);
 		blockingSubscribe(initializeUser());
 		blockingSubscribe(initializeBroker(appBrokerProperties));
+		prepareCLI();
 	}
 
 	void setUpForBrokerUpdate(BrokerProperties brokerProperties) {
@@ -157,6 +176,8 @@ abstract class CloudFoundryAcceptanceTest {
 			"spring.cloud.openservicebroker.catalog.services[0].name=" + appServiceName(),
 			"spring.cloud.openservicebroker.catalog.services[0].description=A service that deploys a backing app",
 			"spring.cloud.openservicebroker.catalog.services[0].bindable=true",
+			"spring.cloud.openservicebroker.catalog.services[0].metadata.properties.serviceInstanceLogsEndpoint=" +
+				getServiceInstanceLogsEndpoint(),
 			"spring.cloud.openservicebroker.catalog.services[0].plans[0].id=" + PLAN_ID,
 			"spring.cloud.openservicebroker.catalog.services[0].plans[0].name=standard",
 			"spring.cloud.openservicebroker.catalog.services[0].plans[0].bindable=true",
@@ -177,6 +198,10 @@ abstract class CloudFoundryAcceptanceTest {
 		appBrokerProperties.addAll(Arrays.asList(openServiceBrokerProperties));
 		appBrokerProperties.addAll(brokerProperties.getProperties());
 		return appBrokerProperties;
+	}
+
+	private String getServiceInstanceLogsEndpoint() {
+		return "https://" + testBrokerAppName() + "." + cloudFoundryProperties.getApiHost().substring(4) + "/logs/";
 	}
 
 	@BeforeEach
@@ -332,7 +357,7 @@ abstract class CloudFoundryAcceptanceTest {
 			.block();
 	}
 
-	private Mono<ServiceInstance> getServiceInstanceMono(String serviceInstanceName) {
+	Mono<ServiceInstance> getServiceInstanceMono(String serviceInstanceName) {
 		return userCloudFoundryService.getServiceInstance(serviceInstanceName);
 	}
 
@@ -423,7 +448,8 @@ abstract class CloudFoundryAcceptanceTest {
 					.getApplicationRoute(testBrokerAppName())
 					.flatMap(appRoute ->
 						webClient.get()
-							.uri(URI.create(appRoute + "/" + operation + "/" + serviceName + "/" + planName + "/" + serviceInstanceId))
+							.uri(URI.create(
+								appRoute + "/" + operation + "/" + serviceName + "/" + planName + "/" + serviceInstanceId))
 							.retrieve()
 							.toEntity(String.class)
 							.map(HttpEntity::getBody)));
@@ -451,11 +477,87 @@ abstract class CloudFoundryAcceptanceTest {
 
 	protected Mono<List<ApplicationDetail>> getApplications(String app1, String app2) {
 		return Flux.merge(cloudFoundryService.getApplication(app1),
-			cloudFoundryService.getApplication(app2))
+				cloudFoundryService.getApplication(app2))
 			.parallel()
 			.runOn(Schedulers.parallel())
 			.sequential()
 			.collectList();
+	}
+
+	private void prepareCLI() {
+		try {
+			cfHome = Files.createTempDirectory("app-broker-acceptance-tests").toString();
+
+			callCLICommand(List.of("cf", "login", "-a",
+				cloudFoundryProperties.getApiHost(),
+				"--skip-ssl-validation", "-u",
+				cloudFoundryProperties.getUsername(),
+				"-p",
+				cloudFoundryProperties.getPassword(),
+				"-o", "test-instances"))
+				.block(Duration.ofSeconds(60));
+			callCLICommand(List.of("cf", "install-plugin", "-f", "-r", "Cf-Community", "Service Instance Logging"))
+				.block(Duration.ofSeconds(60));
+		}
+		catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	protected Mono<String> callCLICommand(List<String> command) {
+		return Mono.fromCallable(() -> {
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Executing command: {}", command);
+			}
+			ProcessBuilder processBuilder = new ProcessBuilder(command);
+			processBuilder.environment().put("CF_HOME", cfHome);
+			processBuilder.redirectErrorStream(true);
+			Process process = processBuilder.start();
+
+			return processOutput(process);
+		}).subscribeOn(Schedulers.boundedElastic());
+	}
+
+	private static String processOutput(Process process) throws InterruptedException, ExecutionException {
+		StringBuffer outputBuilder = new StringBuffer();
+		ExecutorService executor = Executors.newSingleThreadExecutor();
+		Future<Void> future = appendLines(executor, process, outputBuilder);
+		try {
+			future.get(30, TimeUnit.SECONDS);
+		}
+		catch (TimeoutException e) {
+			LOG.info("Process reading timed out after 30 seconds");
+		}
+		finally {
+			future.cancel(true);
+			executor.shutdownNow();
+			process.destroyForcibly();
+			try {
+				process.waitFor(5, TimeUnit.SECONDS);
+			}
+			catch (InterruptedException e) {
+				LOG.warn("Interrupted while waiting for process to terminate", e);
+			}
+		}
+		return outputBuilder.toString();
+	}
+
+	private static Future<Void> appendLines(ExecutorService executor, Process process, StringBuffer outputBuilder) {
+		return executor.submit(() -> {
+			try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+				String line = "";
+				while (line != null) {
+					line = reader.readLine();
+					if (line != null) {
+						outputBuilder.append(line).append('\n');
+						if (LOG.isDebugEnabled()) {
+							LOG.debug("Read line: {}", line);
+						}
+					}
+				}
+			}
+			return null;
+		});
 	}
 
 }
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
@@ -56,7 +56,6 @@ class LoggingRecentAcceptanceTest extends CloudFoundryAcceptanceTest {
 	@AppBrokerTestProperties({
 		"spring.cloud.appbroker.services[0].service-name=" + APP_SERVICE_NAME,
 		"spring.cloud.appbroker.services[0].plan-name=" + PLAN_NAME,
-
 		"spring.cloud.appbroker.services[0].apps[0].name=" + APP_CREATE_1,
 		"spring.cloud.appbroker.services[0].apps[0].path=" + BACKING_APP_PATH,
 		"spring.cloud.appbroker.services[0].target.name=SpacePerServiceInstance",
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2002-2020 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.appbroker.acceptance;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+
+import org.cloudfoundry.operations.applications.ApplicationSummary;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class LoggingRecentAcceptanceTest extends CloudFoundryAcceptanceTest {
+
+	private static final String APP_CREATE_1 = "app-logging-recent-1";
+
+	private static final String SI_NAME = "si-logging-recent";
+
+	private static final String SUFFIX = "logging-recent-instance";
+
+	private static final String APP_SERVICE_NAME = "app-service-" + SUFFIX;
+
+	private static final String BACKING_SERVICE_NAME = "backing-service-" + SUFFIX;
+
+	@Override
+	protected String testSuffix() {
+		return SUFFIX;
+	}
+
+	@Override
+	protected String appServiceName() {
+		return APP_SERVICE_NAME;
+	}
+
+	@Override
+	protected String backingServiceName() {
+		return BACKING_SERVICE_NAME;
+	}
+
+	@Test
+	@AppBrokerTestProperties({
+		"spring.cloud.appbroker.services[0].service-name=" + APP_SERVICE_NAME,
+		"spring.cloud.appbroker.services[0].plan-name=" + PLAN_NAME,
+
+		"spring.cloud.appbroker.services[0].apps[0].name=" + APP_CREATE_1,
+		"spring.cloud.appbroker.services[0].apps[0].path=" + BACKING_APP_PATH,
+		"spring.cloud.appbroker.services[0].target.name=SpacePerServiceInstance",
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=cflinuxfs4"
+	})
+	void shouldReturnBackingApplicationLogs() {
+		createServiceInstance(SI_NAME);
+
+		String lines = callCLICommand(
+			List.of("cf", "service-logs", SI_NAME, "--recent", "--skip-ssl-validation"))
+			.block(Duration.ofSeconds(35));
+
+		assertThat(lines).isNotEmpty();
+
+		assertThat(lines).contains("Created app with guid");
+		assertThat(lines).contains("Updated app with guid");
+		assertThat(lines).contains("APP/PROC/WEB");
+
+		deleteServiceInstance(SI_NAME);
+
+		Optional<ApplicationSummary> backingApplication1AfterDeletion = getApplicationSummary(APP_CREATE_1);
+		assertThat(backingApplication1AfterDeletion).isEmpty();
+	}
+
+}
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
@@ -18,9 +18,8 @@ package org.springframework.cloud.appbroker.acceptance;
 
 import java.time.Duration;
 import java.util.List;
-import java.util.Optional;
 
-import org.cloudfoundry.operations.applications.ApplicationSummary;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -73,11 +72,11 @@ class LoggingRecentAcceptanceTest extends CloudFoundryAcceptanceTest {
 		assertThat(lines).contains("Created app with guid");
 		assertThat(lines).contains("Updated app with guid");
 		assertThat(lines).contains("APP/PROC/WEB");
+	}
 
+	@AfterEach
+	void tearDown() {
 		deleteServiceInstance(SI_NAME);
-
-		Optional<ApplicationSummary> backingApplication1AfterDeletion = getApplicationSummary(APP_CREATE_1);
-		assertThat(backingApplication1AfterDeletion).isEmpty();
 	}
 
 }
--- a/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
+++ b/spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2002-2020 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.appbroker.acceptance;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+import org.cloudfoundry.operations.applications.ApplicationSummary;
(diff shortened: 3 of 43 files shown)


kvmw (approved):

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -451,11 +477,87 @@ private WebClient getSslIgnoringWebClient() {
...
+				.block(Duration.ofSeconds(60));
+		}
+		catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	protected Mono<String> callCLICommand(List<String> command) {
+		return Mono.fromCallable(() -> {
+			if (LOG.isDebugEnabled()) {
```
spikymonkey, line 509: I doubt we're saving anything with this

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -451,11 +477,87 @@ private WebClient getSslIgnoringWebClient() {
...
+			future.get(30, TimeUnit.SECONDS);
+		}
+		catch (TimeoutException e) {
+			LOG.info("Process reading timed out after 30 seconds");
+		}
+		finally {
+			future.cancel(true);
+			executor.shutdownNow();
+			process.destroyForcibly();
+			try {
```
spikymonkey: Is there really any value in this?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
```diff
@@ -451,11 +477,87 @@ private WebClient getSslIgnoringWebClient() {
...
+
+	private static Future<Void> appendLines(ExecutorService executor, Process process, StringBuffer outputBuilder) {
+		return executor.submit(() -> {
+			try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+				String line = "";
+				while (line != null) {
+					line = reader.readLine();
+					if (line != null) {
+						outputBuilder.append(line).append('\n');
+						if (LOG.isDebugEnabled()) {
```
spikymonkey, line 547: I don't think we're saving anything by doing this. Would be useful if there's some significant computation within the `if`, but otherwise I think this pattern just adds clutter.

```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
```diff
@@ -0,0 +1,84 @@
...
+	@Override
+	protected String backingServiceName() {
+		return BACKING_SERVICE_NAME;
+	}
+
+	@Test
+	@AppBrokerTestProperties({
+		"spring.cloud.appbroker.services[0].service-name=" + APP_SERVICE_NAME,
+		"spring.cloud.appbroker.services[0].plan-name=" + PLAN_NAME,
+
```
spikymonkey:
```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
```diff
@@ -0,0 +1,84 @@
...
+		"spring.cloud.appbroker.deployer.cloudfoundry.properties.stack=cflinuxfs4"
+	})
+	void shouldReturnBackingApplicationLogs() {
+		createServiceInstance(SI_NAME);
+
+		String lines = callCLICommand(
+			List.of("cf", "service-logs", SI_NAME, "--recent", "--skip-ssl-validation"))
+			.block(Duration.ofSeconds(35));
+
+		assertThat(lines).isNotEmpty();
```
spikymonkey, line 70: Implicitly handled by the following assertions, no?

```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+		assertThat(lines).contains("Created app with guid");
+		assertThat(lines).contains("Updated app with guid");
+		assertThat(lines).contains("APP/PROC/WEB");
+
+		deleteServiceInstance(SI_NAME);
```
spikymonkey, line 79: This feels like a separate test?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
```diff
@@ -0,0 +1,103 @@
...
+	@Override
+	protected String backingServiceName() {
+		return BACKING_SERVICE_NAME;
+	}
+
+	@Test
+	@AppBrokerTestProperties({
+		"spring.cloud.appbroker.services[0].service-name=" + APP_SERVICE_NAME,
+		"spring.cloud.appbroker.services[0].plan-name=" + PLAN_NAME,
+
```
spikymonkey:
```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
```diff
@@ -0,0 +1,103 @@
...
+			List.of("cf", "service-logs", SI_NAME, "--skip-ssl-validation")
+		)).subscribeOn(Schedulers.boundedElastic());
+
+		StepVerifier.create(createServiceInstanceMono
+				.then(Mono.delay(Duration.ofSeconds(20)).then())
+				.then(getServiceInstanceMono(SI_NAME).retry(5))
+				.then(Mono.zip(logStreamingMono, logStreamingMono))
+			)
+			.assertNext(tuple -> {
+				assertThat(tuple.getT1()).isNotEmpty();
```
spikymonkey, line 83:
```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
```diff
@@ -0,0 +1,103 @@
...
+				assertThat(tuple.getT1()).contains("Connected, tailing logs for service instance");
+				assertThat(tuple.getT1()).contains("[STG/0]");
+				assertThat(tuple.getT1()).doesNotContain("websocket: close");
+
+				assertThat(tuple.getT2()).isNotEmpty();
```
spikymonkey, line 88:
```suggestion
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
```diff
@@ -0,0 +1,103 @@
...
+				assertThat(tuple.getT2()).contains("Connected, tailing logs for service instance");
+				assertThat(tuple.getT2()).contains("[STG/0]");
+				assertThat(tuple.getT2()).doesNotContain("websocket: close");
+			})
+			.verifyComplete();
+
+		deleteServiceInstance(SI_NAME);
+
+		Optional<ApplicationSummary> backingApplication1AfterDeletion = getApplicationSummary(APP_CREATE_1);
```
spikymonkey: As before - it's not clear to me why this is here

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/fixtures/cf/CloudFoundryService.java
```diff
@@ -304,7 +306,7 @@ public Mono<OrganizationSummary> getOrCreateDefaultOrg() {
 	public Mono<List<String>> getSpaces() {
 		return cloudFoundryOperations.spaces().list()
 			.doOnComplete(() -> LOG.info("Success listing spaces"))
-			.doOnError(e -> LOG.error(String.format("Error listing spaces. error=%s" + e.getMessage()), e))
+			.doOnError(e -> LOG.error(String.format("Error listing spaces. error={}" + e.getMessage()), e))
```
spikymonkey:
```suggestion
			.doOnError(e -> LOG.error(String.format("Error listing spaces. error=%s", e.getMessage()), e))
```

spring-cloud-app-broker-autoconfigure/src/test/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceLogStreamAutoConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+			ServiceInstanceLogStreamAutoConfiguration.class
+		))
+		.withPropertyValues(
+			"spring.cloud.appbroker.deployer.cloudfoundry.api-host=https://api.example.local",
+			"spring.cloud.appbroker.deployer.cloudfoundry.username=user",
+			"spring.cloud.appbroker.deployer.cloudfoundry.password=secret"
+		);
+
+	@Test
+	void servicesAreNotCreatedWithoutLoggingOnClasspath() {
```
spikymonkey: Double-negative does not match the test assertions?

spring-cloud-app-broker-autoconfigure/src/test/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceLogStreamAutoConfigurationTest.java
```diff
@@ -0,0 +1,96 @@
...
+			.run(context -> assertThat(context)
+				.doesNotHaveBean(StreamingLogWebSocketHandler.class)
+				.doesNotHaveBean(WebSocketHandlerAdapter.class)
+				.doesNotHaveBean(HandlerMapping.class)
+				.doesNotHaveBean(LogStreamPublisher.class)
+				.doesNotHaveBean(ApplicationLogStreamPublisher.class));
+	}
+
+	@Test
+	void servicesAreNotCreatedWithoutRequiredBeansOnClasspath() {
```
spikymonkey: Double-negative does not match the test assertions?

spring-cloud-app-broker-autoconfigure/src/test/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceRecentLogsAutoConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+			ServiceInstanceRecentLogsAutoConfiguration.class
+		))
+		.withPropertyValues(
+			"spring.cloud.appbroker.deployer.cloudfoundry.api-host=https://api.example.local",
+			"spring.cloud.appbroker.deployer.cloudfoundry.username=user",
+			"spring.cloud.appbroker.deployer.cloudfoundry.password=secret"
+		);
+
+	@Test
+	void servicesAreNotCreatedWithoutLoggingOnClasspath() {
```
spikymonkey, line 49: @Albertoimpl can we come up with some different names for these tests? The double-negative 'not created without' is confusing, and I'm not sure about 'classpath' either since they are asserting on the beans / app context, not the classpath.

spring-cloud-app-broker-docs/src/docs/asciidoc/service-instance-logging.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+Then, include the `serviceInstanceLogsEndpoint` in the catalog metadata properties:
+
+====
+[source,yaml,subs="+quotes"]
+----
+spring:
+    cloud:
+        openservicebroker:
+          catalog:
```
spikymonkey: Inconsistent indentation style

spring-cloud-app-broker-docs/src/docs/asciidoc/service-instance-logging.adoc
```diff
@@ -0,0 +1,46 @@
+[[service-instance-logging]]
+:acceptance-tests-dir: ../../../spring-cloud-app-broker-acceptance-tests/src/main/java/org/springframework/cloud/appbroker/acceptance/
+== Service Instance Logging
+
+You can configure the service broker to be compatible with the https://github.com/pivotal-cf/service-instance-logs-cli-plugin[Service instance logs CLI plugin] in order to tail and stream the logs of the backing application.
+
+
+If you use Gradle, include the following in your application's `build.gradle` file:
```
spikymonkey, line 8: Do we need some instructions for Maven users too?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/recent/endpoint/MultipartEncoder.java
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.appbroker.logging.recent.endpoint;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+class MultipartEncoder {
```
spikymonkey: Can we implement `Closeable`?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
```diff
@@ -0,0 +1,147 @@
...
+				this.stopPublishing(serviceInstanceId);
+				return;
+
+			default:
+				throw new IllegalArgumentException("Unknown operation: " + event.getOperation());
+		}
+	}
+
+	private void startPublishing(String serviceInstanceId) {
+		synchronized (registry) {
```
spikymonkey: I think this could be handled with `ConcurrentMap.compute` rather than synchronising on the whole map instance. We're only ever updating the value for a single key here, so why lock access to the whole map?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
```diff
@@ -0,0 +1,147 @@
...
+			registry.put(serviceInstanceId, new Registration(subscription));
+		}
+	}
+
+	private void stopPublishing(String serviceInstanceId) {
+		if (LOG.isDebugEnabled()) {
+			LOG.debug("Received event to stop publishing logs for {}", serviceInstanceId);
+		}
+
+		synchronized (registry) {
```
spikymonkey: Same again

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
```diff
@@ -0,0 +1,147 @@
...
+			final Registration registration = registry.get(serviceInstanceId);
+			if (registration == null) {
+				if (LOG.isWarnEnabled()) {
```
spikymonkey, line 96:
```suggestion
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
```diff
@@ -0,0 +1,147 @@
...
+						.publishEvent(new ServiceInstanceLogEvent(this, serviceInstanceId, envelope)))
+				.subscribe();
+
+			LOG.debug("Creating new registration for {}", serviceInstanceId);
+			registry.put(serviceInstanceId, new Registration(subscription));
+		}
+	}
+
+	private void stopPublishing(String serviceInstanceId) {
+		if (LOG.isDebugEnabled()) {
```
spikymonkey, line 90:
```suggestion
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
```diff
@@ -0,0 +1,147 @@
...
+		synchronized (registry) {
+			final Registration registration = registry.get(serviceInstanceId);
+			if (registration == null) {
+				if (LOG.isWarnEnabled()) {
+					LOG.warn("Received deregister event for service instance {} but there no event handler registered",
+						serviceInstanceId);
+				}
+			}
+			else if (registration.decrement() == 0) {
+				if (LOG.isDebugEnabled()) {
```
spikymonkey, line 104:
```suggestion
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+										.max()
+										.orElse(lastTimestamp.get());
+
+									lastTimestamp.set(maxTimestamp);
+
+									return Flux.fromIterable(readResponse.getEnvelopes().getBatch())
+										.map(LoggingUtils::convertLogCacheEnvelopeToDropsonde);
+								}))
+							.onErrorResume(error -> {
+								LOG.error("Error during log polling: ", error);
```
spikymonkey:
```suggestion
								LOG.error("Error during log polling", error);
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	protected Flux<Envelope> createApplicationStreamer(String applicationId) {
+		return client.applicationsV2()
+			.get(GetApplicationRequest.builder()
+				.applicationId(applicationId)
+				.build())
+			.map(response -> response.getEntity().getName())
+			.flatMapMany(appName -> {
+				long initialStartTime = Instant.now().minus(5, ChronoUnit.SECONDS).toEpochMilli() * 1_000_000L;
```
spikymonkey, line 74: I think this needs some comments - what is going on here?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+										.map(LoggingUtils::convertLogCacheEnvelopeToDropsonde);
+								}))
+							.onErrorResume(error -> {
+								LOG.error("Error during log polling: ", error);
+								return Flux.empty();
+							});
+
+						return Flux.merge(initialLogs, polledLogs)
+							.retryWhen(Retry.backoff(3, Duration.ofSeconds(5)))
+							.doOnError(error -> LOG.error("Streaming error: ", error));
```
spikymonkey:
```suggestion
							.doOnError(error -> LOG.error("Streaming error", error));
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+									long maxTimestamp = readResponse.getEnvelopes().getBatch().stream()
+										.mapToLong(org.cloudfoundry.logcache.v1.Envelope::getTimestamp)
+										.max()
+										.orElse(lastTimestamp.get());
+
+									lastTimestamp.set(maxTimestamp);
+
+									return Flux.fromIterable(readResponse.getEnvelopes().getBatch())
+										.map(LoggingUtils::convertLogCacheEnvelopeToDropsonde);
+								}))
```
spikymonkey: This is pretty difficult to follow. Could some of these lambdas be extracted into well-named methods?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/endpoint/StreamingLogWebSocketHandler.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class StreamingLogWebSocketHandler implements WebSocketHandler, ApplicationListener<ServiceInstanceLogEvent> {
+
+	private static final Logger LOG = LoggerFactory.getLogger(StreamingLogWebSocketHandler.class);
+
+	private static final UriTemplate LOGGING_URI_TEMPLATE = new UriTemplate("/logs/{serviceInstanceId}/stream");
+
+	private final ApplicationEventPublisher eventPublisher;
+
+	private final ConcurrentHashMap<String, Sinks.Many<Envelope>> envelopeSinks = new ConcurrentHashMap<>();
```
spikymonkey:
```suggestion
	private final ConcurrentMap<String, Sinks.Many<Envelope>> envelopeSinks = new ConcurrentHashMap<>();
```

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/endpoint/StreamingLogWebSocketHandler.java
```diff
@@ -0,0 +1,122 @@
...
+
+		eventPublisher.publishEvent(new StartServiceInstanceLoggingEvent(this, serviceInstanceId));
+		LOG.info("Published event to start streaming logs for service instance with ID {}", serviceInstanceId);
+
+		return session.send(envelopeSink.asFlux()
+				.map(envelope -> session.binaryMessage(
+					dataBufferFactory -> dataBufferFactory.wrap(Envelope.ADAPTER.encode(envelope)))))
+			.doFinally(signalType -> afterConnectionClosed(session, serviceInstanceId))
+			.doOnError(throwable -> LOG.error("Error handling logging stream for service instance {}",
+				serviceInstanceId, throwable));
```
spikymonkey:
```suggestion
			.doOnError(throwable -> LOG.error(String.format("Error handling logging stream for service instance %s",
				serviceInstanceId), throwable));
```

spikymonkey (changes requested):
Thanks @Albertoimpl, this looks like it was a lot of work! I found a few things but nothing major. Happy to chat through any of the comments so we can get this merged in asap 👏

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
Albertoimpl: Did you mean the `if (LOG.isDebugEnabled()) {` ? 
We do have a PMD rule to always guard log calls https://github.com/spring-cloud/spring-cloud-app-broker/blob/e7bd5d6981867895470c4fd8fece4d1dc39a2ba0/src/pmd/pmdRuleSet.xml#L23-L25
I agree that this is not saving much in test code but in production code it's a very common practice to avoid processing resource-consuming `toString` calls.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
Albertoimpl: Yes, we have to wait a few seconds before stopping the process, otherwise, we would not be tailing enough logs to ensure the assertions.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
Albertoimpl: It was a personal preference to know that we got no logs at all when tailing and printing a better error message.

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
Albertoimpl: I have been using the same logic the `log-cache-cli` uses https://github.com/cloudfoundry/log-cache-cli/blob/main/internal/command/tail.go#L134
There is a delay between when the call is made and when it reaches the server, since the official CLI uses 5 seconds, I decided to use the same.
The second call is to ensure we synchronize with what we get from the server as this CLI does.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
Albertoimpl: Should I remove these guards in test code @spikymonkey or leave them for consistency?

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
Albertoimpl: It is more like a cleanup, sadly it is the way we are doing it in all the other tests.

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/endpoint/StreamingLogWebSocketHandler.java
Albertoimpl: Applying it in a different commit to add the import. Thanks!

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
Albertoimpl: You are right. Extracted a few functions, thanks!

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
Albertoimpl: I still prefer having a clear explicit error message https://github.com/spring-cloud/spring-cloud-app-broker/pull/1065#discussion_r1670315061
but happy to remove it if you think it's redundant.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
Albertoimpl: I still prefer having a clear explicit error message https://github.com/spring-cloud/spring-cloud-app-broker/pull/1065#discussion_r1670315061
but happy to remove it if you think it's redundant.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
Albertoimpl: Agree we don't need the assertion. Removing it, but the delete is something we do everywhere else.

spring-cloud-app-broker-autoconfigure/src/test/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceLogStreamAutoConfigurationTest.java
Albertoimpl: What I wanted to say is that if we do not add the required beans to the classpath, meaning the configuration classes, these services should not be created.
How does `servicesAreNotCreatedWhenLoggingIsNotConfigured()` sound?

spring-cloud-app-broker-autoconfigure/src/test/java/org/springframework/cloud/appbroker/autoconfigure/ServiceInstanceRecentLogsAutoConfigurationTest.java
Albertoimpl: Agreed! How about `servicesAreNotCreatedWhenPublisherIsNotConfigured`?

spring-cloud-app-broker-docs/src/docs/asciidoc/service-instance-logging.adoc
Albertoimpl: Agreed! Added it.

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
Albertoimpl: Agreed! Good catch!

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
Albertoimpl: Thanks!

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/recent/endpoint/MultipartEncoder.java
Albertoimpl: I think since `ByteArrayOutputStream` is already `Closeable` it is not necessary but I agree it does not hurt at all. Adding it. Thanks!

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
spikymonkey: That's fair!

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingStreamingAcceptanceTest.java
spikymonkey: If it gives a clearer log message I'm fine with it. Just struck me as odd when I read the test. But I think that makes it a nitpick so not worth spending time on!

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
spikymonkey: That's good context. Maybe a comment and a link in the code? This feels like one of those things we will look back on and wonder why

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
spikymonkey: Can it go in a cleanup method then?

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/LogCacheStreamPublisher.java
Albertoimpl: Agreed, added
```
/***
 * Class to Stream LocCache Envelopes using the same pattern as the
 * <a href="https://github.com/cloudfoundry/log-cache-cli/blob/main/internal/command/tail.go#L134">loc-cache-cli</a>
 */
```

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
Albertoimpl: That's a good point, for some reason I thought we were not able to use that method. Moved these deletions to the `tearDown`

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
spikymonkey: But here we're waiting _after_ the process has been destroyed. If we haven't captured enough logs then doesn't that mean that line 526 is returning too soon? I'm just not clear on what we gain by waiting on line 536 after everything is cancelled / shutdown / destroyed.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
spikymonkey: Ah I didn't realise we had a PMD rule. Yeah I understand the common practice. I feels like overkill to me in tests (and in a lot of cases in production code too, unless we're talking about a hot path, but I realise that's probably a controversial take 😅) but I don't mind the consistency.

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
spikymonkey: I mean... in this case this just adding a redundant check. It's _more_ work.

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/LoggingRecentAcceptanceTest.java
spikymonkey: Nice, thanks!

spring-cloud-app-broker-acceptance-tests/src/test/java/org/springframework/cloud/appbroker/acceptance/CloudFoundryAcceptanceTest.java
Albertoimpl: Oh, apologies I mixed the lines.
This was to give it a few extra seconds to wait for the process to terminate gracefully. I was handling the processes differently when I started writing these ATs but looking at it now, you are very right, there is no need.
Deleting the extra `wait`. Thanks!

spring-cloud-app-broker-logging/src/main/java/org/springframework/cloud/appbroker/logging/streaming/ApplicationLogStreamPublisher.java
Albertoimpl: Actually, the compiler will swallow that and move on. But I agree it's extra noise to the readers' eyes.
Happy to recondiser the removal of the PMD rules but as of this PR, I think we should leave them for consistency.

spikymonkey: Ah you're right of course. And yes let's not get into changing the PMD rules. I forgot we had them.

spikymonkey (approved):

PR #1100, opened 2024-09-23
Adding platform timeout configuration

Albertoimpl:


spring-cloud-app-broker-docs/src/docs/asciidoc/deployment-platforms.adoc
```diff
@@ -21,6 +21,8 @@ spring:
           *client-secret: EXAMPLE_SECRET*
           *default-org: test*
           *default-space: development*
+          *staging-timeout: 15*
+          *startup-timeout: 10*
```
kvmw: I believe this was supposed to be `deployment-timeout`.

kvmw (approved):
I verified the snapshot build and is working  as expected.  Thank you!

The only minor suggestion I have is a message for the failure, If possible. 

Currently, the SI status is shown like the following, in case of timeout error:
```
Showing status of last operation:
   status:    update failed
   message:
   started:   2024-09-26T14:04:45Z
   updated:   2024-09-26T14:04:45Z
```
Maybe having a message like "timeout waiting for deployment" or such can help users without the need to check the logs.

spring-cloud-app-broker-docs/src/docs/asciidoc/deployment-platforms.adoc
Albertoimpl: Good catch, thanks!

Albertoimpl:
Thanks for testing it @kvmw! I like your suggestion regarding error messages but it would require a different set of changes. Mind filing a new issue?

PR #1166, opened 2025-04-03
(2.4.x) Bump to reactor-netty 1.2.4

spikymonkey:
Contains fix for reactor/reactor-netty#3660, which was breaking ATs.


spikymonkey:
Just noticed that we have outstanding Boot patch bumps already queued in PRs. These bring in the required reactor-netty versions, so let's prefer those if they don't bring in any other issues. Just running tests on them now.

